PR #7, opened 2019-08-24
Prepending a linebreak before new words in echo trainer mode

matthiasjordan:
Hi, Willi.
I added some lines that perform a line break before words in echo mode. I.e. before each word that the user is supposed to echo, a line break is done, making the word begin at the start of the line, making it much better readable.
Kind regards
Matthias


matthiasjordan:
The commit 72ef4d6 ("AutoStop mode") introduced auto stop, a mode where the CW generator just beeps w/o printing the characters and stops after each character group (according the limit of the current mode). The user then taps the paddle, the just generated word is printed and after a second paddle tap the next word is generated. This mode works only with char-by-char mode and is intended to help learning hear reading (reading code w/o writing it down).

PR #10, opened 2019-10-25
Proposed fix for scroll buffer update issue

matthiasjordan:
This should fix the weird updating behaviour of the scroll buffer when adding a newline


oe1wkl:
Matthias,

I had commented out your original code change, because the substring function does not work correctly for UTF-8 characters, and therefore created weird results when such a character was present (and sometimes they are, e.g. when using  German Morse code characters…).

What are you trying to do with the loop, outputting one character at a time, instead of outputting the whole string at once?

Willi

matthiasjordan:
Yes, you're right. I noticed that the output is completely broken when having auto stop enabled so this was my attempt at fixing it.

PR #15, opened 2020-02-29
Feature/jw/improve paddle iambic key timing

joewittmer:
This PR contains several commits where each can be viewed separately. Each should improve the signal timing during iambic paddling.

The recommended dit pattern test using audacity to monitor the audio output was performed. The results of the final test with all commits was as follows:

Screen Shot 2020-02-29 at 10 30 59 AM

### DIFF
--- a/Software/src/morse_3_v2.3/morse_3_v2.3.ino
+++ b/Software/src/morse_3_v2.3/morse_3_v2.3.ino
@@ -1346,7 +1346,6 @@ void loop() {
    int t;
 
    //boolean activeOld = active;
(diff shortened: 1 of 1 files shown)


oe1wkl:
The first commit was incorporated into the neweset version (2.4). 2,3 and 4 were not, as the savin is just a few CPU cycles (= microsecond) and hence not really significant. Re. commit #5, I achieved the same with slightly different code.

joewittmer:
Thanks much for taking the time to review this PR, Willi. I'll delete my feature branch.

PR #20, opened 2020-04-30
Fixed few typos

yarda:
Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>


yarda:
I am not sure about the 3.5 mm jack (because I do not have my own Morserino-32 yet), but I guess it's the 3.5 mm.

PR #21, opened 2020-05-08
WiFiTrx: Add DNS support to Wifi TRX mode and use src port 7373

asdil12:
Now a user can use a hostname (eg a dyndns address) as a peer.
Also the source port is now `7373` as well which might help to pass through a NAT.
Also compiling was fixed on case sensitive file systems.

This has not yet been tested with two morserinos. Only with one using wireshark to verify the network traffic!

### DIFF
--- a/Software/src/morse_3_v3.0/MorseWiFi.cpp
+++ b/Software/src/morse_3_v3.0/MorseWiFi.cpp
@@ -22,8 +22,7 @@
 // File file;
 
 WebServer MorseWiFi::server(80);    // Create a webserver object that listens for HTTP request on port 80
-WiFiUDP MorseWiFi::udp;             // Create udp socket for wifi tx
-AsyncUDP MorseWiFi::audp;           // Create async udp socket for wifi rx
+AsyncUDP MorseWiFi::audp;           // Create async udp socket for wifi trx
 
 File MorseWiFi::fsUploadFile;              // a File object to temporarily store the received file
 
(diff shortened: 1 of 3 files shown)


asdil12:
Especial care should be taken when looking at this PR to verify that the sent UDP Packet won't be misunderstood as with the current code the `\0` at the end of the UDP packat is cut of now.
I'm not sure, if this was different with the old `WifiUDP` class.

oe1wkl:
At first look this should not be a problem, as the transmitted data packets should never contain a zero byte (the bit stream cannot contain more than 4 consecutive 0 bits, based on the encoding being used; this would be the sequence dash - end of character - dit (b100001, for example in „mY Fault“). There is also no zero byte expected at the end of a word.

Willi

oe1wkl:
Unfortunately it does not work with two Morserinos. Have not found yet what the issue is. One of them refuses to send, it seems, although it can receive (same behaviour as when you are not connected to the network - touching the paddles doesn't do a thing - the attempt to send seems to block /to wait for something to happen).

asdil12:
Does this still happen when you only enter an ip addres as peer on both morserinos or only when you enter a dns name?

oe1wkl:
Only tested it with IP address, and with broadcast address...

oe1wkl:
Could have been my fault... have to do the test again.... maybe later tonight

oe1wkl:
It was indeed my fault, everything is OK. I also fixed the problem that after WiFi Config the Update / Upload functions didn't work correctly.
Will upload to GitHub tomorrow...and will release a patch version 3.0.1

oe1wkl:
Name resolution is now done when you start the Wifi Trx mode in the menu, after connecting to Wifi, and not every time when you send a packet. Should imporve performance. All unresolvable host names, as well as illegitimate IP addresses are replaced by teh broadcast address. Actual IP address used for teh peer is shown at start of the WiFi Trx mode.
This has been released as version 3.0.1.

asdil12:
The wifi library actually seemt to use some kind of caching as I used tcpdump to look for dns queries on my dns server and there was only a request when sending the first package.
Maybe it will retry after some timeout but it didn't do a dns request for every sent packet.

Issue #31, opened 2020-10-01
Allow connecting to wifi networks with no password


martinmurphy:
Allow the option of connecting to wifi networks with no password.
Some hotel and cafe wifi networks do not have passwords.
While it's not ideal from a security point of view, it would be useful to have the option of connecting.
The current version of the software, uses an empty password in the wifi config form to mean use the current password.
I suggest adding an option to allow an empty password to be set

PR #32, opened 2020-10-01
Allow connection to wifi networks with no password

martinmurphy:
Add an option to the WiFi config to allow connection to networks with no password #31


oe1wkl:
As normally WiFi without password uses a captive portal, this makes no sense - we cannot intercept the captive portal on the Morserino. I do not see any other use case for that.

Issue #30, opened 2020-09-25
support multiple wifi networks


martinmurphy:
I was thinking of adding support for multiple wifi networks
allow setting of up to 3 wifi settings in the config
and allowing choosing of the active config.

This would allow configuring of home network, phone hotspot, and another network, so Morserino-32 could be used at multiple locations easily.

Would you be interested in a PR?

oe1wkl:
Yes, would be useful in some cases.

Willi

oe1wkl:
Has been implemented.

PR #33, opened 2020-10-03
Support multiple wifi network configs

martinmurphy:
Support multiple wifi network configs as mentioned in #30 
Allow details for up to three network to be entered in the wifi config form
Add a new menu option to the wifi functions menu to allow selection of one of the networks

### DIFF
--- a/Software/src/morse_3_v3.0/MorseMenu.cpp
+++ b/Software/src/morse_3_v3.0/MorseMenu.cpp
@@ -21,7 +21,7 @@ using namespace MorseMenu;
 //////// variables and constants for the modus menu
 
 
-const uint8_t menuN = 41;     // no of menu items +1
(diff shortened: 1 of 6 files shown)


martinmurphy:
@oe1wkl I didn't include my other PR (#32) in this.  I can include it if you like.

oe1wkl:
Not sure how useful it would be to allow networks without password- most of these are captive portals that require to enter credentials through a web browser, which wouldn’t work with a Morserino....

Willi

Von meinem Mobiltelefon gesendet

martinmurphy:
When I wrote it, I was staying at a hotel that had wifi with no password, so I needed it.
However, I agree that they are not as common, so I'm not too concerned.
I think  that being able to switch between some predefined networks easily, e.g. home, office, and phone hotspot is more useful.

Martin.

Issue #45, opened 2021-10-22
Add option to auto-select Koch lesson


JosePadron:
- Add new menu selection to define number of times symbol is keyed correctly before moving automatically to the next lesson.
- Select a new Koch lesson after symbol is keyed correctly multiple times without errors.

PR #46, opened 2021-10-22
Add option to auto-select Koch lesson after symbol is keyed correctly multiple times in a row

JosePadron:
Related to https://github.com/oe1wkl/Morserino-32/issues/45

- [x] Add new menu.
- [x] Add logic for auto-select new Koch lesson.
- [x] Bench test.


JosePadron:
> Jose, this will require some testing before I can commit it.

I understand. Let me know if you need anything else from me.

PR #50, opened 2022-02-05
Add adaptive koch echo trainer

DO1GDO:
When i reached koch lesson 25 i noticed that new characters are not repeated as often as e.g. in koch lesson 10. You need to key a lot a characters until the new character is asked again.

This pull-requests adds "Adaptive Random" mode to the Koch Echo Trainer.

The "Adaptive Random" mode modifies the random selection of characters with feedback from the keyed responses. A wrong character will increase its probability to be selected. A correctly keyed character will reduce its probability.

To start the adaptive mode start: Koch Trainer > Echo Trainer > Adapt. Rand.

Remarks:

- Probabilities will be reset every time you start "Adaptive Random" mode.
- The last koch lessons / characters have a higher probability at the beginning of the session.
- At the beginning of the session, every character will be selected once (in random order).
- After every character was selected once, the next characters are selected randomly, characters that have been keyed wrong will have a higher probability to be selected.
- A wrong keyed character will also increase the probability of the character left and right. E.g. "z/?" is asked and you reply with "g/?". Then the probability of z will increased and probability of / will also be increase a little.
- Only the first wrong character will be analyzer. Subsequent input will not be analyzer. E.g. "z/?" is asked and you reply with "gz/?". Probabilities will be increased the same way as in previous example.
- Do not go away. Every timeout for keying will be counted. This will typically destroy the learned probabilities. Exit the adaptive mode and start again.
- Do not expect to have any fun in this mode. The adaptive mode will tease you with the characters that you cannot key 100% correctly every time. Once you have keyed a character wrong, that will give you the chance to key the character wrong again and thus increasing its probability to be selected again. If you reached a total level of frustration, switch back to koch random mode and relax some time before using the "Adaptive Random" mode again.

### DIFF
--- a/Software/src/Version 4/MorseMenu.cpp
+++ b/Software/src/Version 4/MorseMenu.cpp
@@ -21,7 +21,7 @@ using namespace MorseMenu;
 //////// variables and constants for the modus menu
 
 
-const uint8_t menuN = 42;     // no of menu items +1
+const uint8_t menuN = 43;     // no of menu items +1
 
 const String menuText [menuN] = {
   "",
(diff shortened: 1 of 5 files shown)


DO1GDO:
Please let me know if there are some coding guidelines or best practices that i should apply to the code before accepting the pull request. I will be happy to improve the code.

oe1wkl:
I will have a look at this and test it over the next 2 weeks... am a bit busy at the moment. But looks interesting by all means!

PR #53, opened 2022-02-15
WiFi vs. HiFi and image source

chbw:
I stumbled over something that probably is a typo (WiFi vs. HiFi) and fixed the markdown source. The .pdf  files are untouched, but need to be rebuilt.
And the image source in the repository's README.md resulted in a 404, so I tried to fix this too.
73
Christoph


oe1wkl:
Thanks for pointing out the errors!

PR #65, opened 2022-12-29
Added LICW Charset and Carousel offset

GavinAndrews:
Added learning order for LICW and also carousel offset to allow you to jump in when you like...

Offset +1 Week means you joined the carousel one week late so join BC1 at TIN rather than REA; I started at LCD so was Week +3

### DIFF
--- a/Software/src/Version 4/MorsePreferences.cpp
+++ b/Software/src/Version 4/MorsePreferences.cpp
@@ -1332,6 +1332,12 @@ void MorsePreferences::readPreferences(String repository) {
     MorsePreferences::cwacKochSeq = pref.getBool("cwacKochSeq");
     MorsePreferences::lcwoKochSeq = pref.getBool("lcwoKochSeq");
     MorsePreferences::licwKochSeq = pref.getBool("licwKochSeq");
+
+    if ((temp = pref.getUChar("carouselStart")))
+       MorsePreferences::carouselStart = temp;
+    else if (morserino)
+       pref.putUChar("carouselStart", MorsePreferences::carouselStart);
+
     MorsePreferences::useCustomCharSet = pref.getBool("useCustomChar");
     MorsePreferences::customCharSet = pref.getString("customCharSet", "");
     MorsePreferences::quickStart = pref.getBool("quickStart");
--- a/Software/src/Version 4/MorsePreferences.cpp
+++ b/Software/src/Version 4/MorsePreferences.cpp
@@ -64,6 +64,8 @@ Preferences pref;               // use the Preferences library for storing and r
   uint8_t MorsePreferences::kochFilter = 5;                   // constrain output to characters learned according to Koch's method 2 - 50 (45??)
   boolean MorsePreferences::lcwoKochSeq = false;              // if true, replace native sequence with LCWO sequence 
   boolean MorsePreferences::cwacKochSeq = false;              // if true, replace native sequence with CWops CW Academy sequence 
+  boolean MorsePreferences::licwKochSeq = false;              // if true, replace native sequence with LICW Carousel sequence 
+  uint8_t MorsePreferences::carouselStart = 0;                // Week Offset for LICW Carousel
   boolean MorsePreferences::useCustomCharSet = false;         // if true, use Koch trainer with custom character set (imported from file)  
   String  MorsePreferences::customCharSet = "";               // a place to store the custom character set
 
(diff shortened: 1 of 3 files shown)


oe1wkl:
While reviewing you proposed changes re. LICW, I noticed that you introduced <bk>  as a new pro sign. This was not included on purpose, as I could not find any authoritative reference for this being a pro sign - all references point to bk as two separate characters (_…  _._ not _..._._).
I know that LICW seems to treat this as a combined character /true pro sign), but I think this is wrong.
Do you have any reference that would void my point of view?

Willi

N7CQR:
This discussion came up last month on another email list. Here’s an excerpt from my posting for what it’s worth. I concluded that it’s an open ended situation:

on KB6NU’s page about your novice accent (updated from the original) he also lists it as <BK> to distinguish it from’Back-BK’. I dug around in my teaching materials and one thing I use for new CW ops is a template published by’CW 4 U’  on how to complete a typical QSO. Helps them when they’re starting out to make contacts. It also lists prosigns and abbreviations and <BK> is listed as‘fast change’ and under abbreviations BK is ‘break or break in’. 

http://www.cw4u.org/

Dan Presley 503-701-3871
***@***.*** com 
***@***.***

N7CQR:
Here is a copy of the template from CW4U. When I started CW traffic handling back in the late 1960’s it was drilled into us to send <BK> as one character to distinguish from ‘BK’ which could also mean ‘back’,as in ‘back to u’ which is often used at the end of the conversation as opposed to a quick <BK> where you needed to break in for clarification or something like an emergency call. I still use this as it makes more sense to me to separate the two. CW is a constantly evolving language. I can usually date an operator by their use of abbreviations-‘SRI OM NIL CPY PSE AGN <BK> probably started using CW 40-50 years ago. Now I see SMS abbreviations in CW use-‘LOL’ instead of’HI-HI’. Keeps us on our toes. 

http://www.cw4u.org/download/CW4U_poster_ENG.pdf

Dan Presley 503-701-3871
***@***.*** com 
***@***.***

GavinAndrews:
I am brand new to CW and have not progressed beyond BC1 yet so I am not sure.  I know BK is taught as a 'Character' in BC2 and in the practice page it is no different from B followed by K.  The only way I could see to represent that was a BK run together. I think that should be OK with a bit of poetic licence especially since B and K as separate letters are taught in other parts of the carousel.

oe1wkl:
I will include it as suggested. To complete the task, I am also adding the new pro-sign character to the built-in CW decoder. A new version release (4.5) an be expected towards end of January at the latest).

Willi

GavinAndrews:
I think I found a bug... it doesn't seem to remember Carousel Adjustment between power downs.  Must have missed something there... will look.

GavinAndrews:
Seems to load the preference every time now

GavinAndrews:
Changed the Carousel Adjustment to indicate the starting week e.g. Start REA rather than using a non-obvious +3Weeks

oe1wkl:
While going through this in more detail, I found a few things that do not work, and honestly I have not figured out yet HOW they SHOULD work; the difficulty comes predominantly with the carousel offsets, but also with the fact, that the LICW curriculum always seems to restrict the characters in the words and abbreviations to SOME lessons (the last three in B1, ??? In BC2):

- for generating random letters, your approach is clear enough, and it probably does what it would
- it does not work for picking CW Abbrevs and English words
	I went through the LICW course descriptions, and see that for BC1  they pick words that use characters learned in the last three lessons (so you have to attend at least three lessons, and have learned 9 characters, before you start training with words, right?).
	For BC2 they assume you have learned all characters of BC1, plus the current lesson (but when you are in BC lesson 2, also the last lesson?

I try to understand which characters should be included when you set Koch lesson to a specific number, to filter out all those words and abbreviations that should be included at that stage. But so far I haven't figured it out.

As it stands you cannot use the Morserino to train words and abbreviations, as the M32 charset for a given lesson will not be identical with the training plan of LICW.

Can anyone help me to get an understanding of that? I would have loved to roll out the next version of the firmware within January, but now it seems this just will not be possible, until I understand how it SHOULD work.

A simple example where I am at right now:
- I set the charset to LICW
- I set the carousel offset to HOF
- I select Koch lesson 9

When training CW abbreviations, I get e.g. the following:
es sri ant stn sp net sn pse test ii nr

But that’s using the first 9 characters of the LICW charset, starting from the first lesson, and not taking into account the offset. Once you get into BC2, it gets even muddier….

Willi

oe1wkl:
I found more problems:
The settings for the carousel offset do not work correctly for BC2 (there are 8 classes with 3 chars each, and one with 2 chars).
I will try to find a solution for that.

Willi

GavinAndrews:
I tried to reply by email but it didn't seem to work...

Indeed BC1 and BC2 don't have the same cycles... I worked on the basis that BC1 mattered the most but maybe that was too simplistic and as you point out it doesn't make a late (T+6W, T+7W,  T+8W) start possible.  I guess the simplest solution to that is to allow either a choice between a BC1 and BC2 character set and then allow T+0...T+5 if you have chosen BC1 or T+0...T+8W if you have chosen BC2... alternatively allow two offsets... which seems a bit over the top!

When the first 9 chars are chosen and you started with HOF then the only characters available should be HOFUWBREA and I think that works in the echo and single character parts.  Maybe the CW abbrevs and words part needs tweaking and it shouldn't really generate characters other than HOFUWBREA if that's how it works in other character sets.

I'm definitely open to ideas on the late start parameters.

As for practice lessons.... I think it should use as many characters as you have selected rather than limiting this to just one or two past lessons.  As you go around the Carousel you need to add them in and potentially stay with a complete carousel before transitioning to BC2.

My 2p worth anyway.

oe1wkl:
After going through the LICW Carousel documentation again, and sleeping over it, I think I have found a working solution that should address all issues seen so far in this respect.
Right now testing it, if everything turns out ok, I will release a new version within the next 2-3 days.


Willi

oe1wkl:
I#m going to close this request. I have included those parts that were working OK into my code, and replaced other parts with my own code to fix the problems with generating abbrevs and words correctly for each entry point and all classes in BC1 and BC2 of the LICW cirriculum.

PR #66, opened 2022-12-29
Allow Keys that are wired the other way around by honoring polartity

GavinAndrews:
Allow keys that are wired the other way around (like mine!)

### DIFF
--- a/Software/src/Version 4/MorseDecoder.cpp
+++ b/Software/src/Version 4/MorseDecoder.cpp
@@ -19,6 +19,7 @@
 extern unsigned long genTimer;      /// needed for echo trainer 
 extern morserinoMode morseState;
 extern int leftPin;
+extern int rightPin;
 
 //////// methods for class Decoder /////////
 
@@ -42,7 +43,7 @@ void Decoder::setup() {
 }
 
 
-#define straightPin leftPin
+#define straightPin (MorsePreferences::reversePolarity ? rightPin : leftPin)
 
 
 boolean Decoder::checkInput() {         /// check if we have a tone signal at A6 with Gortzel's algorithm or from a key (fromKey == true), and apply some noise blanking as well


oe1wkl:
That's already built in... Von meinem Mobiltelefon gesendetAm 29.12.2022 um 19:38 schrieb GavinAndrews ***@***.***>:﻿Allow keys that are wired the other way around (like mine!)



You can view, comment on, or merge this pull request online at:
  https://github.com/oe1wkl/Morserino-32/pull/66

Commit Summary

  b66eca1  Allow Keys that are wired the other way around by honoring polartity


File Changes  (1 file)

  
    M
    Software/src/Version 4/MorseDecoder.cpp
    (3)
  


Patch Links:

  https://github.com/oe1wkl/Morserino-32/pull/66.patch
  https://github.com/oe1wkl/Morserino-32/pull/66.diff

GavinAndrews:
Can you be a bit more specific?  The code only samples leftPin for straight keys

digitalRead(straightPin)

where 

#define straightPin leftPin

on  a v4 board...

leftPin = 32;
rightPin = 33;

So only 32 is sampled.  My key is wired (via the phono to 33)

oe1wkl:
Ah, I see - my thought was about paddles not straight key.
Your solution is elegant, although I have a bit of a problem with supporting „wrongs“. With straight key connections there are two possible wrongs: the one you address, a wrongly connected stereo plug, and the other wrong is a connection with a mono plug. The latter one has been catered for (mono plugs are justifiable by historical reasons, and most transceivers will be ok with it, too. Your „wrong“ is worse, I think, as I suppose that no commercially available transceiver will work with your wiring…. But as the change is minor (and probably has no unwanted side effects), and elegant, there’s nothing that speaks against it….

Willi

PS: Did not have a close look at the other changes you proposed - times are a bit stressful at the moment…

GavinAndrews:
That's interesting.  I've always considered myself (and my keys) special rather than wrong but it's interesting that my keys are not wired conventionally.  My straight key wiring was due to using a Logikey Model K-3 keyer and I need the key to do dits to allow me to flip to straight key mode using the key alone.

Now you are suggesting most keys are wired the other way around I will propose an alternative... since otherwise people who use paddles the wrong way around might be sad if their Morserino was deaf when they plugged in a straight key...

... there is an obvious solution... OR left and right

GavinAndrews:
Ah... so with a mono plug are you saying that right is always shorted?

If that is the case maybe we cannot do this.

GavinAndrews:
FYI I resolved this by changing the wiring on my existing keys and paddles!

oe1wkl:
👍Von meinem Mobiltelefon gesendetAm 30.12.2022 um 16:13 schrieb GavinAndrews ***@***.***>:﻿
Closed #66.

Issue #78, opened 2023-08-04
Feature request: speed compatibility with CWops


beedaddy:
I would like to use the Morserino for a course at CWops as a practice tool. However, the speed initially used there is 25WPM with a Farnsworth speed,  i.e. an effective speed, of 4WPM. Unfortunately, this cannot be mapped on the Morserino at this time. I would therefore wish that such speeds are also supported. At best by simply entering speed and effective (Farnsworth) speed in the settings.

beedaddy:
The feature came with version 5.0.1. Thanks a lot!

PR #79, opened 2023-08-05
Allowing to set spaces as used in CWops beginner course.

beedaddy:
Allowing to set inter-word and inter-character spaces as used in CWops beginner course. This would be a minimal solution for #78.


oe1wkl:
Will be taken care of in Release 5.0.1, to be published immediately...

PR #92, opened 2024-01-16
Avoid initializing serial port twice; Recognize \r

wcravens:
The Serial port was initialized twice.  Once through an explicit call to Serial.begin() and again from Heltec.begin with SerialEnable = true.   This breaks the serial interface entirely in my build environment and with minicom on OSX (probably the same for Linux).

Also added \r as a recognized ENTER in SerialEvent.

### DIFF
--- a/Software/src/Version 5/m32_v5.ino
+++ b/Software/src/Version 5/m32_v5.ino
@@ -386,8 +386,6 @@ void setup()
    //// 7. check for press of key/paddle at start, to initiate hw config
    //// 8. do the remaining initialisations
 
-  Serial.begin(115200);
-  delay(50); // give me time to bring up serial monitor
   // reserve 200 bytes for the serial inputString variable defiend above:
   inputString.reserve(255);
 
@@ -2719,7 +2717,7 @@ void serialEvent() {
         inputString += inChar;
         // if the incoming character is a newline, set a flag so the main loop can
         // do something about it:
-        if (inChar == '\n') {
+        if (inChar == '\n' || inChar == '\r') {
           stringComplete = true;
           break;
         }


oe1wkl:
What build environment are you using? The binary created with my environment (Arduino) runs perfectly fine with an OSX machine on the other end of the serial port (I am using OSX almost exclusively), and I have not heard that this binary has a problem with the serial port anywhere....

wcravens:
I'm using arduino-cli.  The IDE consistently failed to build.  However,
I've not tried again since I finally got cli working.

I thought about that before I sent the patch, but sent it anyway because
the changes are for the better.

-----
***@***.*** ~ % arduino-cli version
arduino-cli  Version: 0.35.0 Commit: ebab482d Date: 2024-01-02T10:47:07Z

***@***.*** ~ % arduino-cli lib list
Name                    Installed Available      Location
 Description
ArduinoJson             7.0.1     -              LIBRARY_LOCATION_USER -
Heltec ESP32 Dev-Boards 1.0.2     1.1.5          LIBRARY_LOCATION_USER
Library for Heltec ESP32 (or ESP32+Lo...

***@***.*** ~ % arduino-cli core list
ID                 Installed Latest Name
arduino:avr        1.8.6     1.8.6  Arduino AVR Boards
Heltec-esp32:esp32 0.0.6     0.0.6  Heltec ESP32 Series Dev-boards
-----

wcravens:
What terminal client do you use?  I'm using Minicom 2.8 from homebrew.

I've just reflashed my m32 to the m32_v5.1.ino.wifi_lora_32_V2.bin and can
confirm that I get no response when I enter the "PUT device/protocol/on"
command.  However, it all works as expected after the patch.

Wes

oe1wkl:
I use both the built-in terminal in the Arduino IDE, and also USB serial with Chrome, when using tegmento.org (which supports the M32 protocol nicely)

wcravens:
I investigated further.  And yes the Arduino IDE defaults to 'Newline'
(drop down to the right of the input box in the serial monitor).  The
tegmento is also explicitly sending a newline '\n'.  And I can confirm that
the 'production' firmware does work when '\n' is sent.

But, this is not the default for traditional terminals. \n is a line feed.
\r is a carriage return.  \r is what is traditionally sent in the terminal
when the enter, ⏎, return key is pressed.  The serial device 'usually'
responds with a '\n' line feed which moves the terminal to the next line.

Please reject the pull request if you don't agree with this change.

PR #120, opened 2025-03-01
Headphone audio bug fix

naiiawah:
Fix an issue with Headphone detection when plugged in at boot.

We found an issue with the Headphones being plugged in at bootup or sleep recovery not being detected. Found a mention in the TI support forum that the audio had to be enabled first for the detection to work.

Refactored the soundEventHandler so that a call to enable the main speaker audio could be called directly to enable the audio deck and then call the event handler 2nd. The test for Headphones would work now when the handler was called.


oe1wkl:
Thanks, Brian!

W

PR #143, opened 2026-04-20
Separate internal paddle from external key

dramaturg:
This separates the internal Paddle from the externally connected key.

I am learning with a straight key and found it pretty annoying to have to switch back and forth all the time when I want to use the internal paddles on the couch.

* The external key and its settings stay the same
* This adds a new "Paddle Mode" setting to control how the built in paddle behaves

73 de Seb, DM3SEB

### DIFF
--- a/Software/src/Version 6 and newer/MorseDecoder.cpp
+++ b/Software/src/Version 6 and newer/MorseDecoder.cpp
@@ -52,11 +52,15 @@ boolean Decoder::checkInput() {         /// check if we have a tone signal at A6
 ///// straight key is connected to external paddle connector (tip), i.e. the same as the left pin (dit normally)
 
 if (fromKey) {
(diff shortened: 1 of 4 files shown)


oe1wkl:
I am not yet ready to merge this, there are a few issues I need to address first. 
The most important one is that now you have two preferences, "Keyer Mode" (with options Iambic A, Iambic B, Ultimatic, Non-Squeeze and Straight Key), and "Paddle Mode" (with options Iambic A, Iambic B, Ultimatic and Non-Squeeze ). This is going to confuse users - many of them have external paddles, and I cannot believe that any of them will have the internal paddles in a different mode than the external ones. The only meaningful use case is an external straight key. So I'd rather have a separate Preference "External Key Mode" with options "As built-in" and "Straight Key".
It might be helpful to ask on the user group what users think about such a feature.

PR #174, opened 2026-05-24
Fix: TFT-Pocket dark on USB-only power (revert PR #157 source path)

oe1wkl:
## Summary

PR #157 ("Replace DisplayWrapper with direct LovyanGFX on TFT") regressed the M32 Pocket display on **USB-only-powered devices**:
- Panel stays dark
- Backlight is on
- CPU / encoder / audio / keyer all functional
- Battery-fitted Pockets are unaffected

Confirmed on Device B (no battery, USB only) — MAC `58:E6:C5:5A:E0:34`. Different USB supplies show different brief flashes at boot, then darkness, consistent with a marginal power-rail / panel-init interaction.

## Bisect

| Commit | Behaviour |
|---|---|
| `9cbeb13` V8.0 source | works ✅ |
| `b8de2c5` (#156, immediately before #157) | works ✅ |
| `1b36e44` (#157) | **dark** ❌ |
| current master | dark ❌ |
| current master + this revert | works ✅ |

## Attempted fix-forward (none worked)

Each tried on Device B with the post-#157 in-tree `LGFX` class:
- Pin LovyanGFX to `~1.1.x`
- Move `LGFX display` to its own translation unit
- Defer all panel/bus/light config from constructor to `init()`
- Re-init the display after `wifiWarmup` (the biggest current spike)
- Disable all `MorseGameMode` / `RadioCave` / `Morsel` warmups
- 200 ms delays after every spike point in `setup()`
- `cfg.bus_shared = false`
- SPI clock 40 MHz → 10 MHz
- Remove the `void display() {}` no-op shadow
- Full V8.0-style wrapper (`LGFX_Impl` + forwarding `LGFX` + `static lcd` in a dedicated TU, then again in a project-local `lib/M32PocketLGFX/`)

Only restoring `DisplayWrapper` as the actual `lib_dep` — i.e. the literal V8.0 source path — keeps the panel alive on USB-only Device B.

## Likely root cause (not 100% confirmed)

C++ static-init order. The verbose PIO link line shows:

```
[before --start-group]:
  LovyanGFX/*.o (individual)
  src/*.o        (project source, including MorseOutput.cpp.o)
[inside --start-group]:
  libESP32Encoder.a … libDisplayWrapper.a … libBLE.a
```

`DisplayWrapper.cpp.o` lives inside `libDisplayWrapper.a`, **inside the start-group**. The in-tree `M32PocketLGFX.cpp.o`, even when moved into `lib/M32PocketLGFX/`, was linked as a plain `.o` before the start-group, alongside project src. Different position → different `.init_array` slot → different static-init order. The plausible cascade is through LovyanGFX's lazy `static Panel_NULL nullobj` inside `LGFX_Device::setPanel(nullptr)` — triggered at base-class ctor time, so position-dependent.

That's a hypothesis, not a proof. Follow-up work in [issue/branch TODO] will try to force PIO to bundle `lib/M32PocketLGFX/` into a real archive and verify the link position matches DisplayWrapper's. If that works, PR #157's cleanup can stand.

## Changes

Reverts the source-level part of #157. Keeps every other post-#157 master improvement (Morsel, 8-bpp sprites, multiplayer, …).

- `MorseOutput.cpp`: `LGFX display` → `DisplayWrapper display`
- `MorseGameMode.cpp`, `MorseGame.cpp`, `MorsePileup.cpp`, `MorseRadioCave.cpp`, `MorseMorsel.cpp`: include + extern type back
- `platformio.ini`: re-add `https://github.com/oe1wkl/DisplayWrapper`

`M32PocketLGFX.h` and `IntelOneMono*.h` are left on disk (dead code) so a future fix can revisit the direct-LGFX approach without restoring all the scaffolding.

## Test plan

- [x] Build clean on `pocketwroom`
- [x] Flash on Device B (no battery, USB only) — panel lights up, splash + menu render normally
- [ ] Flash on Device A (battery installed) — confirm no regression
- [ ] Quick smoke: Keyer, Generator, Echo, Koch, Trx, Decoder, one game (Morsel or Invaders)

### DIFF
--- a/Software/src/Version 6 and newer/MorseGame.cpp
+++ b/Software/src/Version 6 and newer/MorseGame.cpp
@@ -21,7 +21,7 @@ char gameCharBuffer = 0;
 #include "MorseMenu.h"
 #include "morsedefs.h"
 #include "ClickButton.h"
-#include "M32PocketLGFX.h"
+#include "DisplayWrapper.h"
 
 #include <LovyanGFX.hpp>
 #include <ESP32Encoder.h>
@@ -37,7 +37,7 @@ extern void serialEvent();
 extern boolean checkPaddles();
 extern boolean doPaddleIambic(boolean, boolean);
 extern boolean leftKey, rightKey;
-extern LGFX display;
+extern DisplayWrapper display;
 extern ESP32Encoder rotaryEncoder;
 extern const int PinCLK;
 extern const int PinDT;
--- a/Software/src/Version 6 and newer/MorseGameMode.cpp
+++ b/Software/src/Version 6 and newer/MorseGameMode.cpp
@@ -29,7 +29,7 @@
 #include <Arduino.h>
 #include <ESP32Encoder.h>
 
-#include "M32PocketLGFX.h"
+#include "DisplayWrapper.h"
 #include "MorseOutput.h"
 #include "MorsePreferences.h"
 #include "morsedefs.h"
@@ -98,7 +98,7 @@ namespace {
   // the failure path of allocate() so that callers see a coherent menu
   // regardless of whether enter*() succeeded.
   void restoreMenuDisplay(bool leftHanded) {
-    (&display)->setRotation(leftHanded ? 0 : 2);
+    DisplayWrapper::getLGFX()->setRotation(leftHanded ? 0 : 2);
     MorseOutput::initDisplay();
     MorseOutput::setTheme(MorsePreferences::pliste[posTheme].value);
     pinMode(PinCLK, INPUT_PULLUP);
@@ -115,7 +115,7 @@ namespace {
   LGFX_Sprite *allocate(int rotation, bool leftHanded, uint8_t colorDepth) {
     lastLeftHanded = leftHanded;
 
-    auto *lcd = (&display);
+    auto *lcd = DisplayWrapper::getLGFX();
     lcd->setRotation(rotation);
     lcd->fillScreen(TFT_BLACK);
 
@@ -175,7 +175,7 @@ void MorseGameMode::warmup() {
   // inside a game session. (See module header comment for why timing
   // matters: the lazy alloc otherwise lands inside what would become the
   // game sprite's tail region.)
-  auto *lcd = (&display);
+  auto *lcd = DisplayWrapper::getLGFX();
   LGFX_Sprite tmp(lcd);
   tmp.setColorDepth(16);
   if (!tmp.createSprite(lcd->width(), lcd->height())) return;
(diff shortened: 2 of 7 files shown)


oe1wkl:
## Follow-up: fix-forward attempted, root cause remains unidentified

Spent another round trying to keep PR #157's `M32PocketLGFX` direct-LGFX approach while restoring the panel on USB-only Pockets. **All attempts produced a dark screen.**

What I tried on a separate branch (now deleted — no commits worth keeping):

1. **Move `LGFX display;` into a project-local `lib/M32PocketLGFX/`** so the static-init slot lands in a library archive position (matching where V8.0's `static LGFX lcd;` inside `libDisplayWrapper.a` sits). PIO needed `"build": { "libArchive": true }` in `library.json` to actually emit `libM32PocketLGFX.a`. Confirmed in the link line:
   - Master (broken): `M32PocketLGFX.cpp.o` linked individually before `--start-group`.
   - Fix-forward: `libM32PocketLGFX.a` linked inside `--start-group`, exactly like `libDisplayWrapper.a`.
   - **Still dark.** → Archive position alone is not the cause.

2. **Full V8.0-style wrapper (`LGFX_Impl` base + `LGFX` thin forwarder + `static LGFX_Impl lcd;` inside the archive)** — structurally as close to `DisplayWrapper` as possible without literally being it. Updated `MorseGameMode.cpp` to use `display.getLGFX()` instead of `(&display)`. **Still dark.**

3. **Both libraries present in the link** — fix-forward setup as above, AND `oe1wkl/DisplayWrapper` re-added to `lib_deps` (unused by source, so its `static LGFX lcd;` constructs but never gets `begin()`'d). **Still dark.** Mere presence of `libDisplayWrapper.a` in the link line doesn't change the outcome — what matters is that DisplayWrapper's `lcd` is the one driving the panel.

### Falsified hypotheses

- LovyanGFX library version drift (1.1.16 vs 1.2.21)
- Static-init order from archive-vs-object link position
- Two-LGFX-instance interference
- SPI clock rate / `bus_shared` setting
- Current-spike clustering in `setup()` (200 ms delays everywhere didn't help)
- Wrapper structure (composition vs inheritance from `LGFX_Device`)
- `display()` no-op shadow swallowing flushes

### What's left

Whatever specifically about `oe1wkl/DisplayWrapper` (vs my structurally-equivalent in-tree wrapper) keeps the panel alive on USB-only-powered Pockets is not in any code-level difference I can identify. The empirical picture is "only the literal `oe1wkl/DisplayWrapper` library works."

Likely needs an oscilloscope on the panel's VDD/IOVDD/RST/CS rails during boot to see what hardware-level difference the two builds produce. Parking the investigation here — the revert in this PR is the working ship-now answer.

PR #189, opened 2026-06-18
Ble midi

khkremer:
**Description:**
  
  This PR adds a BLE MIDI output feature that allows the Morserino-32 to act as a wireless MIDI key adapter. When enabled, pressing and releasing the left or right paddle sends standard MIDI Note On / Note Off
  messages over Bluetooth Low Energy to any connected host - a phone, tablet, PC, or SDR - without a USB cable.

  Use case: this makes the M32 usable as a wireless CW key input for software that accepts MIDI input, like SDR software. The functionality will be the same as with the TinyMIDI hardware sold by N6ARA.

**What was added:**
  
  - MorseMidi.cpp / MorseMidi.h - a BLE GATT service using the standard MIDI-over-BLE spec UUIDs, compatible with iOS (Core MIDI), macOS, and Windows (with a BLE MIDI driver). Runs as a FreeRTOS task;
  re-advertises after a disconnect with a short delay.
  - Four new preferences under Key/Paddles/Keyer:
    - BLE MIDI Out - on/off (default off)
    - MIDI Channel - 1–16 (default 1)
    - MIDI Left Note - 0–127 (default 20)
    - MIDI Right Note - 0–127 (default 21)
  - Paddle events wired in checkPaddles(): each debounced left/right transition fires noteOn/noteOff with the configured note number.
  - BLE MIDI is mutually exclusive with the existing BLE keyboard feature - enabling one disables the other (they share the same hardware).
  - CONFIG_BLE_MIDI compile-time flag added to both hardware variant environments in platformio.ini (Classic M32 and M32 Pocket).
  - User manual updated in English and German: four new preference rows, new Appendix 10 with connection instructions, glossary entries.

**What was not changed:** the existing BLE keyboard feature, all CW training modes, keyer behaviour, and audio are unaffected. BLE MIDI adds no overhead when disabled.

**Testing:** tested with an M32 Pocket connecting to iOS via a yet unreleased app that interacts with a HermesLite2 transceiver and midimittr; left and right paddle events arrive as independent simultaneous-capable Note On/Off messages, suitable for iambic-aware software.


oe1wkl:
Thanks, Karl Heinz — nice feature, and I'm glad it's a runtime preference rather than a compile-time choice.

One design change I'd like before we merge: instead of a separate **BLE MIDI Out** on/off preference alongside **BLT Kbd Output**, fold MIDI into a single **BLE Output** selector (`Nothing / Vband Keying / Decoded / Vband+Decoded / Generic Kbd / MIDI`).

Reasoning: the HID keyboard and MIDI share the single ESP32 BLE radio, so they are mutually exclusive. With two independent toggles a user can turn both on, and the current guard only prevents MIDI from starting while the keyboard runs — not the other way round. Enabling both and then entering the keyer can therefore bring up two BLE stacks, and tearing one down (`BLEDevice::deinit(true)`) frees the radio out from under the other. A single selector makes the "both on" state impossible by construction and keeps the choice entirely in the preferences.

I've drafted this refactor on top of your branch (builds clean for both the Classic V2 and the Pocket) — happy to push it to a branch or send you a patch, whichever you prefer. The only follow-up is a small manual update, since the four MIDI rows collapse into the selector option plus the three MIDI sub-settings.

khkremer:
I was trying to get away with just the minimum necessary understanding of how Morserino works to implement my change :) Your proposed way of implenenting the preferences is definitely the way to do it. Can you please push it to a branch so I can pull the changes. 

Thank you!

oe1wkl:
Great — pushed it to a branch on this repo: **`ble-output-single-selector`**

https://github.com/oe1wkl/Morserino-32/tree/ble-output-single-selector

It's your `BLE_MIDI` plus a single commit with the refactor, so it should come in as a fast-forward. From your clone:

```
# add this repo as a remote once, if you don't have it:
# git remote add oe1wkl https://github.com/oe1wkl/Morserino-32.git
git fetch oe1wkl
git checkout BLE_MIDI
git merge oe1wkl/ble-output-single-selector
```

(or just cherry-pick the one commit, `cd7e417`). Builds clean for both the Classic V2 and the Pocket.

The only remaining piece is the EN+DE manual update to match — the four MIDI rows collapse into the single **BLE Output** selector option plus the three MIDI sub-settings.

oe1wkl:
I'd appreciate if you can test the proposed changes against 'real' MIDI hardware. If all goes well I'll integrate it into the main repository.WilliVon meinem Mobiltelefon gesendetAm 18.06.2026 um 17:15 schrieb Karl Heinz Kremer ***@***.***>:﻿khkremer left a comment (oe1wkl/Morserino-32#189)
I was trying to get away with just the minimum necessary understanding of how Morserino works to implement my change :) Your proposed way of implenenting the preferences is definitely the way to do it. Can you please push it to a branch so I can pull the changes.
Thank you!

khkremer:
I will try my best to fit some testing in, but I have a busy weekend and a busy week coming up. This is our club's hamfest weekend.

oe1wkl:
No worries, we are not in a hurry...

khkremer:
I finally had a chance to verify that your changes work in my environment. I was also discussing this with a friend, and we came up with another way of doing this: There is a reason to stay compatible with the TinyMIDI hardware, this makes it easy to just use the Morserino instead of the TinyMIDI device. However, when doing that, the Morserino will show garbage on the screen, because my keying is actually interpreted on the remote end, and that sounds different than what I am seeing being decoded on the display. Going forward, it may make sense to have two different MIDI modes: Once that emulates TinyMIDI, and a second one that actually uses the Morserino as the keyer and just sends note on/off messages that will be interpreted on the receiving side as straight key messages - all the hard stuff is already taken care of in the Morserino. This would require a 2nd entry for the "BLE Output" menu: MIDI (Keyer) and MIDI (Paddles) [I need to figure out how many characters we can display in that menu] - the latter one would just send the raw left/right paddle information and the first one would utilize the Morserino as keyer. Thoughts?

oe1wkl:
Sounds good to me. But is there not a problem with latency when you send just the paddle information over BLE? Paddle timing usually is extremely time-sensitive… But I have no experience with the TinyMidi device, which apparently also just sends the paddle information….

Willi

deqw:
This is just the option I need to have the bluetooth keyboard option working out of the box with Vband on MacOS. Can't wait ;o) Thanks!

Edit: Hi @oe1wkl @khkremer I was able to generate the binary with the midi branche. The Morserino is recognised fine as midi device `system_profiler SPBluetoothDataType | grep -i midi
          Morserino32 MIDI:`

But no midi arriving in Safari (yet). Probably a developer setting somewhere, looking into it.

Screenshot 2026-06-25 at 14 13 24

khkremer:
Did you connect to the BLE device using the "Audio Midi Setup" on the Mac? To do that, bring up the "Audio Midi Setup" application and then select to "Show MIDI Studio":

image

Then, find the Bluetooth icon and select the BLE device and finally click on the "Connect" button:

image

I am not familiar with MIDI in Safari. What page are you on that supports that?

deqw:
Yep, had that one already running. In your first post you said `Use case: this makes the M32 usable as a wireless CW key input for software that accepts MIDI, such as VBand`. So I made the assumption that this would work with the Vband channel webpages. Can you explain a bit more please?

Screenshot 2026-06-25 at 16 38 40

khkremer:
Sorry, I think this was a copy and paste error. Vband should not be there. I just edited the comment. 

Vband keyboard is supported by the Morserino, but macOS wants to verify what type of keyboard you have. Maybe the key to make this work is to add functionality that allows the Morserino to send the key presses that are required for the Bluetooth keyboard to be paired.

deqw:
@khkremer OK I see, thanks for correcting that.

When I tried v8.1 at first I was able to use Vband Keying with text fields or an editor that accept keyboard input. But with the Vband website channels that did not work (maybe on Windows it does). But now, after using your branche, Vband Keying doesn't work anymore. I already compiled v8.1 master again but that did not help. I'm going to upload the release now to see what that brings.

> Maybe the key to make this work is to add functionality that allows the Morserino to send the key presses that are required for the Bluetooth keyboard to be paired.

Do you mean like sending out the `[` and `] ` characters to Vband? Maybe @oe1wkl can chime in to give his opinion  about this? Thanks.

oe1wkl:
Morserino is sending [ and ] when it is set up for VBand keying. BLT is active as soon as you start CW Keyer, and you can pair it with your PC or Mac (just connect to it, it identifies as Morserino). It uses US Keyboard codes, so by default it should work, and I guess a number of people have it working with VBand.
Maybe ask a question at the Morserino groups.io user group

Willi

deqw:
@oe1wkl Thanks for follow up.  I'm back on v8.1 master release again on a M32(v2). When I use Vband Keying on Vband webpage nothing is coming in, when I use Vband+Decoded I can fetch the characters in a text editor. I can use my keyboard to get `[ `and `]` to Vband, but not via the M32.

deqw:
Hallo @oe1wkl,

Are these `[` and `]` also supposed to pass through `void bluetoothTypeCharacter(const char chr);` ?

I did some debugging and Vband keying never goes through the BT keyboard. When adding Decoded I see output like from my printf:

> Hello: bluetoothTypeStringHello: bluetoothTypeCharacterBT CHAR: chr=101 'e'  usage=0x08 mod=0x00
REPORT: mod=0x00 key=0x08
Hello: bluetoothTypeStringHello: bluetoothTypeCharacterBT CHAR: chr=32 ' '  usage=0x2C mod=0x00
REPORT: mod=0x00 key=0x2C

Edit: never mind, created a pull request.

oe1wkl:
@khkremer: do you want to change anything on your PR? you mentioned two types of BT midi modes yoiu would like to have.

khkremer:
@oe1wkl - Sorry for the delay, I got sidetracked by non-hobby tasks and had to put this project on the back burner. I will pick this up again soon. I have the 2nd BT mode running, but it needs more testing. I want to merge in your latest changes, and then run another round of tests.

oe1wkl:
Any news here? If you prefer (or just won#t find the time to do it), I can try and pick this up wher it stands...

khkremer:
Sorry for the delay. I did pull your v9 code and merged my code into that - including the audio snippets. I am more and more convinced that my original approach sending two different MIDI notes is not the best way to handle this, and that just having the one MIDI note controlled by the keyer output is the way to go. I may rip the original code out and just keep the new iteration in.

Issue #140, opened 2026-03-22
Feature Request: Act as a Vail adapter


bebrown:
The [Vail Adapter](https://vailadapter.com) allows a Morse key to be connected to a computer so that it can work with compatible websites and apps.

Essentially, the Vail Adapter acts as a virtual keyboard with two keys: `[` and `]`.  It sends `[` for dit and `]` for dah. Some websites expect other keys (for example, `z`/`x`, `.`/`/`); having the keys be a configurable option would be a plus.

The Morserino implementation could act as a virtual USB keyboard or use Bluetooth to send the characters.

oe1wkl:
Why is the current Bluetooth keyboard functionality of the Morserino not good enough?WVon meinem Mobiltelefon gesendetAm 22.03.2026 um 07:34 schrieb Barry Brown ***@***.***>:﻿bebrown left a comment (oe1wkl/Morserino-32#140)
Some websites expect other key combinations, such as z/x or .//. Having the keys sent as a configuration option would be a plus.

oe1wkl:
The Morserino just sends one CTRL key code - all sites like VBand must be set to straight key mode (the Morserino does the Iambic Keying, and the audio - which also avoids any delays).

PR #191, opened 2026-06-26
Fix VBand BT brackets bug

deqw:
As [ and ] were not coming through at all via Vband Keying some changes were necessary. This is working now with https://hamradio.solutions/vband/ 

In Vband settings just choose Iambic Keyer mode when that is the way you work with Morserino.

I see no timing problems: the sound from the Morserino and the PC are at the same time.

Tested with Morserino-32 V2 on MacOS 26.5.1.

### DIFF
--- a/Software/src/Version 6 and newer/MorseBluetooth.cpp
+++ b/Software/src/Version 6 and newer/MorseBluetooth.cpp
@@ -252,7 +252,7 @@ void MorseBluetooth::bluetoothTypeCharacter(const char chr)
 		uint8_t val = (uint8_t)chr;
 		if (val > KEYMAP_SIZE)
 			return; // character not available on keyboard - skip
-		KEYMAP map = keymap[chr];
+		KEYMAP map = keymap[(uint8_t)chr];
 
 		// create input report
 		InputReport report = {
@@ -273,6 +273,12 @@ void MorseBluetooth::bluetoothTypeCharacter(const char chr)
 		input->setValue((uint8_t *)&NO_KEY_PRESSED, sizeof(NO_KEY_PRESSED));
 		input->notify();
 	}
+}
+
+void MorseBluetooth::bluetoothTypeVbandElement(char c)
+{
+	if (isBleConnected)
+		bluetoothTypeCharacter(c);
 }
 
 // Emit a Shift+Enter HID report (soft return in word processors / chat apps).
--- a/Software/src/Version 6 and newer/MorseBluetooth.h
+++ b/Software/src/Version 6 and newer/MorseBluetooth.h
@@ -25,6 +25,7 @@ namespace MorseBluetooth
 	void stopBluetooth(void);
 	void bluetoothTypeLCTRL(bool ctrl);
 	void bluetoothTypeCharacter(const char chr);
+	void bluetoothTypeVbandElement(char c);
 	void bluetoothTypeString(const String& str);
 };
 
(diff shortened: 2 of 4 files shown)


oe1wkl:
Actually, I don't think that there was a bug, and the proposed PR changes the behaviour significantly. The original design was to operate VBand in straight key mode. The reason for this was that we use the Morserino as keyer, including generating the sidetone, to avoid any delays as they were reported by some. The character that has been used for keying VBand was the left CTRL key (any of the CTRL keys as well as [ and ] can be used for VBand, and for straight key mode it does not matter whioch is used).
I just tested it on VBand (on a Mac, and another ham tested it on MorseInvaders on a PC), and it works as designed.
The proposed change would alter the behaviour (I am not saying that it would not work, but it would work differently).
Therefore I do not see a reason to change anything.

deqw:
Well then we don't call it a bug.

The fact that your solution actually is not working correctly on macOS was already signalled by someone else (@bebrown) in March this year. https://github.com/oe1wkl/Morserino-32/issues/140#issuecomment-4107376004

And @bebrown already had a very good explanation for why it does not work: 

> However, my computer isn't detecting the right-control key from the Morserino. It's probably because the Mac needs to go through the keyboard setup assistant, which needs certain keys to be pressed on the keyboard to identify them.

All I did was bring to your attention again (see also #189) that it does not work as you think it does. I debugged your solution: CTRL keys are not arriving in the software with reason mentioned before and again: the keyboard cannot be validated by macOS because you cannot press the keys because there are no keys to press...

Instead of spending your time on proving that your solution works in your situation you better had tested my fix and then come to conclusions. Maybe resulting in some additional coding, maybe not. Now you don't know. The only conclusion now is: "it would work differently". Well that is what Github is for. Your conclusion to ignore the issue and close it this way is just strange.

Bye.

Screenshot 2026-06-29 at 18 25 33

oe1wkl:
But it HAS been tested on MacOS! Von meinem Mobiltelefon gesendetAm 29.06.2026 um 18:59 schrieb deqw ***@***.***>:﻿deqw left a comment (oe1wkl/Morserino-32#191)
Well then we don't call it a bug.
The fact that your solution actually is not working correctly on macOS was already signalled by someone else ***@***.***) in March this year. #140 (comment)
And @bebrown already had a very good explanation for why it does not work:

However, my computer isn't detecting the right-control key from the Morserino. It's probably because the Mac needs to go through the keyboard setup assistant, which needs certain keys to be pressed on the keyboard to identify them.

All I did was bring to your attention again (see also #189) that it does not work as you think it does. I debugged your solution: CTRL keys are not arriving in the software with reason mentioned before and again: the keyboard cannot be validated by macOS because you cannot press the keys because there are no keys to press...
Instead of spending your time on proving that your solution works in your situation you better had tested my fix and then come to conclusions. Maybe resulting in some additional coding, maybe not. Now you don't know. The only conclusion now is: "it would work differently". Well that is what Github is for. Your conclusion to ignore the issue and close it this way is just strange.
Bye.
Screenshot.2026-06-29.at.18.25.33.png (view on web)

deqw:
While looking into this again and wasting some additional hours I finally got this working after changing some code in MorseBluetooth.cpp. But when I then uploaded the published 8.1 release again to test it was still working. So along the road something has changed on my Mac. But what?

It was probably the HID report map length I experimented with while debugging that forced macOS to reload the HID descriptor. But when I then used the same 8.1 release with my other mac it didn't work with that one. So how to get that one working too?

In general: the problem I and some others ran into was that macOS was rejecting the HID reports caused by a cached HID descriptor, probably from some earlier connection with the device. Flushing that cache results in macOS using the correct descriptor and CTRL key comes through again.

To get rid of that cache you can go something like: Forget the device - Turn Bluetooth OFF - Wait some time and Turn Bluetooth ON - Restart the Mac - Pair the device again. Anyway that was the procedure I used to get the other Mac going.

Maybe some useful information for others that run into this situation.

PR #194, opened 2026-07-03
Ble serial protocol

morria:
# BLE Serial: the M32 serial protocol over Bluetooth LE

## What this adds

A new **BLE Serial** feature (`CONFIG_BLE_SERIAL`, both canonical variants): the existing line-based M32 serial protocol — unchanged, still protocol 1.3 — becomes available over a Nordic-UART-Service BLE GATT server, in addition to USB. Any BLE-capable app can connect (no pairing, one central at a time), send `PUT device/protocol/on`, and then remote-control the Morserino exactly like a USB client, including `PUT cw/play/...` to key CW. The motivating use case is iOS, where apps cannot open a classic serial-over-Bluetooth connection at all; a companion Messages-style iOS app — [**CW Keyer**](https://github.com/morria/CW-Keyer) — has been built against this and verified against real hardware (its README documents the transport contract from the client's side).

- Controlled by the existing Bluetooth preference, now renamed **Bluetooth Use** with one added option: *No Bluetooth / VBand Kbd / Decoded output / VBand+Decoded / Generic Kbd / **BLT Serial Prot.***. Selecting **BLT Serial Prot.** hands the BLE stack to the serial protocol; the device then advertises as **Morserino-32** from the next return to the main menu. The new option is deliberately last (value 5), so existing stored settings (0–4) keep their meaning — verified on hardware. The 6th option and `maximum = 5` exist only in `CONFIG_BLE_SERIAL` builds.
- Because it is one selector, the keyboard and the serial protocol are **structurally** exclusive — one selector value owns the BLE stack, no separate preference, no runtime arbitration.
- USB and BLE sessions are independent and can run simultaneously; each transport has its own `protocol/on|off`. A never-handshaken USB terminal stays byte-silent, exactly as today.
- Any WiFi activity suspends BLE Serial for that session (shared 2.4 GHz radio); it resumes automatically at the top menu.

| Morserino-32 | CW Keyer (iOS) |
|:---:|:---:|
| Morserino-32 | CW Keyer app |

## How it's built (details in `devdocs/ble-serial/`)

- **Emission vs delivery:** protocol output is generated under `protocolActive()` (either transport handshaken) and delivered per transport by a small tee (`M32ProtocolOut`); the protocol engine (`serialDecode`/`m32Get`/`m32Put`) is fully shared, one command line per poll on both transports. Without `CONFIG_BLE_SERIAL` the tee degenerates to plain `Serial` (flag-off build is byte-identical to master, verified).
- **New module `MorseBleSerial`** + pure, host-tested `BleByteRing.h` (SPSC rings with producer-latched session marks, poisoned-overflow semantics so a torn line can never execute, notify flow control with single-writer monotonic counters). 17 host tests, runnable with plain clang++ (`devdocs/ble-serial/test_blebytering.cpp`).
- **Own BLE identity:** BLE Serial advertises from a static random address. This matters: hosts that ever bonded to the Bluetooth keyboard (VBand) auto-connect to the chip's public address the instant it advertises and silently capture the device. Found on hardware; the distinct identity keeps both features working.
- Advertising uses direct GAP calls: `BLEAdvertising::start()` + `setScanResponse(true)` with a 128-bit UUID builds an oversized scan response and silently never starts advertising (library bug, only visible at debug log levels the project can't enable).

## Review + verification provenance

- Design went through a 3-round adversarial review (maintainer/hardware/architect personas) before implementation; the implementation diff through a 5-dimension review with two independent verification votes per finding — 12 confirmed findings, all fixed (the worst: a disconnect replayed the whole session's commands; regression-tested).
- **Verified on a real M32 Pocket:** scan/connect (MTU 515), handshake, chunked multi-KB reassembly, remote keyer start, audible `cw/play`, batched writes, per-transport `protocol/off` in every casing, un-handshaken silence, re-handshake, disconnect → re-advertise → reconnect cycles — the full 8-step scripted suite passes on the current selector build. Test: `devdocs/ble-serial/ble_m32_test.py` (Python + bleak).
- Builds: `heltec_wifi_lora_32_V2` RAM 31.3 % / flash 60.7 % (master: 29.6 % / 60.5 %), `pocketwroom` 39.2 % / 67.7 % (37.4 % / 67.5 %); flag-off `pocketwroom` at master baseline. Merged up to `V8.2-beta.1`; both variants build clean on the merged tree.

## Remaining before merge (hardware checklist, `devdocs/ble-serial/DESIGN.md`)

- [ ] WiFi suspend/resume ×5 per board with free-heap deltas (library leaks GATT objects per `deinit(false)`/`init()` cycle — measure)
- [ ] Mid-game multiplayer entry drops BLE safely; suspended-session auto-sleep; remote self-disable
- [ ] Classic V2: scripted steps 1–8, BT-keyboard interplay, LoRa+BLE coexistence, WiFi-upload-after-BLE (heap regime), ≥35 WpM stalled-client CW-timing soak
- [ ] Snapshot store/recall with the Bluetooth Use setting; factory reset

## Relationship to the single-selector direction (#189)

The original version of this PR added a *separate* BLE Serial on/off preference. Per review feedback, it has been **folded into the single "Bluetooth Use" selector** (option 5) — so there is now one Bluetooth setting, not two competing ones, matching the direction discussed on #189. Options 0–4 keep their historic order and values so stored settings survive the change.

## Notes for the maintainer

1. **`MorseBluetooth::stopBluetooth()` uses `deinit(false)` when `CONFIG_BLE_SERIAL` is compiled in** (flag-off builds keep today's `deinit(true)`, byte-identical). Rationale: `deinit(true)` leaves the ESP32-BLE library's `initialized` flag set, making any later `BLEDevice::init()` a silent no-op and the next `createServer()` a permanent hang, so BLE Serial could never start after the keyboard stopped. While tracing that, I noticed the keyer→menu→keyer keyboard re-init path may be affected by the same library behavior on current master — I could be misreading the flow, so I've written up exactly what I observed in `devdocs/ble-serial/DESIGN.md` for you to judge (and I'm glad to help test either way).
2. **Rule-9 hardening:** added two `static_assert`s pinning `prefName[]` and `pliste[]` to the `prefPos` enum count — a missing or extra entry in either array is now a compile error on every variant instead of a silent shift of all following preferences. (It caught a leftover entry during the selector rework, so it has already earned its keep.)
3. **Display width:** the mapped value `"BLT Serial Prot."` is 16 characters, over the classic OLED's 14-char value line, so it truncates there. Easy to shorten (e.g. "BLT Serial", "M32 Protocol") — happy to do whichever you prefer.
4. `Documentation/Protocol Description/M32 Protocol.md` gained a **Transports** section (protocol version stays 1.3 — a new transport, no new commands, per the additive-change rule); its PDF needs regenerating, which I didn't do for lack of a build script. The user-manual PDFs *were* rebuilt (EN + DE, through `normalize_ids.lua`) and carry both the V8.2 content and the Bluetooth Use changes. FAQ, `UX_CONVENTIONS.md` (§10.1 background connectivity services) and CLAUDE.md (new hard rule: protocol output via `m32out`/`protocolActive()`) updated alongside.

## Scope, review, and AI disclosure

This is a large change for one PR — apologies for that; the transport, the gating sweep and the docs are hard to land separately without a broken intermediate state. The commits are staged to be reviewed in order (flag+tee → gating sweep → preference → BLE module → dispatch → lifecycle → review fixes → hardware fixes → selector rework), each building green on both variants. If you'd prefer it split, restructured, or rebased onto other in-flight work, just say so.

Developed with Claude Code (as the commit trailers show), reviewed and hardware-tested by me: the scripted end-to-end suite above ran against my own M32 Pocket, CW audibly keyed from a phone and a Mac. I'll respond to review myself and can iterate quickly — the remaining checklist items are queued on my bench.


oe1wkl:
Hi Andrew! 
This is a cool idea, thanks for that and all the work you (and Claude ;-) put into it.
But I still do not like the idea of having two conflicting BLE preferences, even when their use case is very different. I'd rather rename the BLE preference from "BLT Kbd Output" to "Bluetooth Use", (and probably some of the option names together with it - "No Bluetooth", "BLT Serial Prot.", "VBand Kbd", "Decoded output", "VBand+Decoded", "Generic Kbd" - need to be careful about the oder here in order not to break the behaviour of existing settings; so BLT Serial probably would need to go to the end of the line). So if you could rework it along these lines, I would greatly appreciate it!
Willi

morria:
Hi Willi,

Thanks for the quick look — reworked as suggested, and I agree it's cleaner this way.

**BLT Kbd Output** is now **Bluetooth Use** with the options *No Bluetooth / VBand Kbd / Decoded output / VBand+Decoded / Generic Kbd / BLT Serial Prot.* — the new option deliberately at the end (value 5), so stored settings keep their meaning. I verified that concretely on my Pocket: it had value 2 stored from before the change, and after flashing it still read "Decoded output". The separate BLE Serial preference is gone entirely (including its snapshot special-casing), and the keyboard-vs-serial exclusion is now just the selector — no extra logic, no notice screen. The 6th option only appears in builds with `CONFIG_BLE_SERIAL`. Manuals (EN+DE, PDFs rebuilt), FAQ and the protocol description are updated to match, and the full BLE test suite passes on hardware with the new selector.

While reworking this I also added two `static_assert`s that keep `prefName[]` and `pliste[]` aligned with the `prefPos` enum — a missing or extra entry in either array is now a compile error on every variant instead of a silent shift of all following preferences. (It promptly caught a leftover entry in my own rework, so it has already earned its keep.)

One small thing to check with you: the option label **"BLT Serial Prot."** is 16 characters, which is over the 14-character value line on the classic OLED, so it will truncate there. Happy to shorten it — "BLT Serial" or "M32 Protocol" would fit — or leave it as is if you prefer the explicit name.

73, Andrew (W2ASM)

oe1wkl:
Hi Andrew, 

first of all: thank you for your work. Folding the feature into the
single "Bluetooth Use" selector is exactly the direction I wanted after #189.
I've built the PR on top of current master for both canonical variants (clean,
sizes match your numbers) and run it through an in-depth review here (with
Claude Code - Fable 5 - assistance on my side). The architecture is sound; what follows are
concrete issues the review confirmed, all with small, localized fixes.

### Please fix before merge

1. **Stale `isBleConnected` can key the torn-down HID characteristic.**
   `stopBluetooth()` (`MorseBluetooth.cpp:215ff`) never clears `isBleConnected`,
   and selector value 5 (0b101) passes the historic HID bitmask gates
   (`(value & 0x1)==1` at `m32_v6.ino:2711`/`3432`, `(value & 0x6)>=2` at the
   decoded-output sites). Sequence: VBand keyboard session with host connected
   → menu (stop) → switch selector to BLT Serial Prot. → BLE Serial re-inits
   the stack → enter CW Keyer → every paddle press runs
   `input->setValue()/notify()` on the pre-deinit HID characteristic against
   the re-owned Bluedroid instance. This was unreachable under `deinit(true)`;
   your `deinit(false)` change makes it live. Your own BLE-serial code even
   documents the root cause ("never trust onDisconnect to be delivered through
   deinit") — the HID side needs the same treatment. Fix: clear
   `isBleConnected` in `stopBluetooth()`, and ideally exclude value 5 from the
   HID bitmask gates explicitly.

2. **`PUT snapshot/recall` bypasses the selector change-switch.**
   `m32_v6.ino:4174` calls `doReadSnapshot()` without `writePreferences`, and
   `posBluetoothOut` is not snapshot-exempt. A BLE client recalling a snapshot
   stored with selector 0–4 leaves BLE Serial running while the selector no
   longer reads 5; the next keyer entry then starts the HID `bluetoothTask` on
   top of the live NUS stack, and `bleSerialGattsHook`
   (`MorseBleSerial.cpp:111`) swallows the keyboard's `ESP_GATTS_REG_EVT`,
   clobbering `ourGattsIf` — breaking flow control and the module's own stated
   mutual-exclusion invariant. The menu recall path is fine (it goes through
   `writePreferences`). Fix: run the same posBluetoothOut side-effect in the
   PUT recall path.

3. **`txEnqueue()` admits bytes during a backoff episode.**
   `MorseBleSerial.cpp:347` calls `produceSome()` *before* the `txBackoff`
   check, contradicting the stated "drop everything until fully drained"
   policy: since the tee writes one byte per call, the client receives spliced
   fragments of multiple JSON objects, and sustained echo traffic can starve
   the `used()==0` clear condition indefinitely. Fix: hoist the backoff check
   above the first `produceSome()`.

4. **The 20 ms self-drain runs inside the CW keying path.**
   Speed/volume encoder turns emit `jsonControl` from within keyer mode
   (`m32_v6.ino:2677/2696`); with a congested client and a full TX ring,
   `txEnqueue`'s drain deadline (`MorseBleSerial.cpp:352ff`) stalls the loop
   task for up to 20 ms — double the ~10 ms delays that were audible,
   field-reported CW-timing bugs we have seen previously. 
   Please don't drain synchronously from`txEnqueue` at all: enqueue what 
   fits and let `pump()` drain.

5. **Session-control messages leak across transports.** With both USB and BLE
   handshaken, `PUT device/protocol/off` on one transport sends
   `end m32protocol` / "Goodbye!" to *both* clients (`m32_v6.ino:4069` and
   `3863` emit through the tee before clearing the flag), so the other client
   sees a false end-of-session while its session on the device stays active.
   Same mechanism in reverse: the `BLE LINE TOO LONG` / `BLE RX OVERFLOW`
   errors (`m32_v6.ino:3848/3878`) are injected into a handshaken USB client
   that never provoked them. Session-scoped messages need targeted
   single-transport delivery, not the tee.

6. **`sessionResetPending` is cleared at the end of the reset block**
   (`MorseBleSerial.cpp:294`), so an edge signalled by the Bluedroid task
   while the block executes is lost (classic lost-update; stale lines from a
   dead session can then execute). Free fix: clear-first, like `takeLineReset()`.

7. **Downgrade hazard: stored selector value 5 clamps to 4 = Generic Kbd** on
   any build without `CONFIG_BLE_SERIAL` (`readPreferences` constrain, pliste
   max 4). Flashing an older release would silently turn a BLE-Serial user
   into a BLE keyboard typing decoded CW at any bonded host. Please map
   out-of-range `posBluetoothOut` to 0 instead. (This is probably a minor issue
   rarely encountered by normal users, but might bite developers).

### Non-blocking, but please consider

- **Display string:** yes, please shorten — "BLE Serial" is my preference. Note
  it's slightly worse than plain truncation today: `displayValueLine`'s
  padding arithmetic (`maxLength - valueLine.length()`) wraps unsigned for
  >14-char values, appending the whole `emptyLine` (cosmetic ghosting on the
  classic OLED, no overrun). A `max(0, …)` there would harden it regardless.
- Your new UX_CONVENTIONS §10.1 promises an on-device splash when WiFi
  suspends BLE Serial; the code emits only the protocol message. Either add
  the splash or amend the convention text — I'd prefer the splash.
- `bleSerialEvent()` mirrors `serialEvent()` with real divergences already
  (USB `put device/protocol/OFF` with uppercase OFF fails in `m32Put` — value
  isn't lowercased — while BLE accepts it; 400-byte cap BLE-only). A shared
  line-assembler would remove the drift risk; at minimum please fix the USB
  case-sensitivity while you're in there.
- `new BLE2902()` per `init()` leaks per suspend/resume cycle — a static
  instance like your `serverCallbacks`/`rxCallbacks` pattern fixes it in one line.
- The raw scan response hardcodes `"Morserino-32"` plus hand-computed length
  bytes; derive from `BLE_NAME`/`strlen` so the two can't diverge.
- The tee delivers `serializeJson` output one byte per `write()` — for
  multi-KB GET responses that's thousands of call chains in the same loop pass
  as the keyer. A buffered write (measureJson + one `write(buf, n)`) would help.
- ~5.6 KB of static ring/staging RAM is paid even when BLE Serial is never
  selected. Not a blocker, but worth a thought (heap-allocate in `init()`?).

### On your notes

1. **`deinit(false)`:** I follow the rationale and accept it for the shared
   stack — but it now also applies to keyboard-only users, permanently
   withholding the BTDM memory after any keyboard session, on the classic V2
   too. I'm therefore treating your open checklist item (WiFi upload / OTA
   after a BT session on the V2, with free-heap deltas) as a hard merge gate.
   The keyer→menu→keyer re-init observation you wrote up: good catch, I'll
   look at it on master independently of this PR.
2. The `static_assert`s on `prefName[]`/`pliste[]`: thank you, that one has
   bitten me before — happy to have it pinned at compile time.
3. No need to split the PR; the staged commits review fine as they are.

The remaining hardware checklist in `DESIGN.md` (WiFi suspend/resume heap
deltas, classic V2 suite, snapshot/factory-reset interplay) stands as the
merge gate together with the fixes above. Once those land I'm glad to run the
classic-V2 tests on my bench as well.

Thanks again — this is one of the best-prepared PRs this project has received.

vy 73, Willi OE1WKL

oe1wkl:
@morria — following up on your note 1: I looked into the keyer→menu→keyer
re-init path, and your reading of the flow was correct. The keyboard has been
one-shot per boot since the `ESP.restart()` after "Stop BT Kbd" was removed in
March: `deinit(true)` leaves the library's `initialized` flag set, so the next
session's `init()` no-ops and `createServer()` blocks forever in `registerApp`,
exactly as you described in DESIGN.md.

This is now fixed on master (06a3502, c61ac37), verified on hardware:
`stopBluetooth()` uses `deinit(false)` unconditionally, clears
`isBleConnected`, and a teardown flag skips `onDisconnect`'s 5 s reconnect
grace so mode exit stays fast. For this PR that means your flag-gated
`stopBluetooth()` hunk is redundant — a rebase onto current master should
resolve it trivially (and the flag-off build now behaves like flag-on there).
Everything else should be unaffected.

Thanks for the careful write-up — it made the verification straightforward.

morria:
Thank you Willi. I'm out of tokens until the 8th, so I'll clean up the PR then.

morria:
Hi Willi,

thank you for the thorough review — and for chasing the keyer→menu→keyer re-init down to its root on master. I've merged current master into the branch (the flag-gated `stopBluetooth()` hunk resolved to your unconditional version, as you predicted) and addressed everything. All seven blocking items plus the requested cleanups are in `6c84e76`; both canonical variants build clean and the `BleByteRing` host tests pass.

### The seven fixes

1. **Stale HID keying** — master's `isBleConnected` clear covers the root cause; on top of that there is now `MorseBluetooth::keyboardMode()`, which returns 0 for selector value 5, and *every* HID bitmask gate goes through it (keyTransmitter, keyOut, the three decoded-output sites, both keyer-entry checks, and `bluetoothTypeString`'s own `>= 0x4` test). 5 = 0b101 can no longer pass any keyboard gate, connected flag or not.
2. **`PUT snapshot/recall`** — the recall path in `m32Put` now runs the same side effect as the `writePreferences` change-switch: recalled selector ≠ 5 stops BLE Serial (after the `ok`, which still reaches the BLE client via the flush inside `stopWithNotice`). Recall *to* 5 is started by the top-menu backstop as usual.
3. + 4. **`txEnqueue`** — went with your stronger suggestion: no synchronous drain at all. Backoff is checked *before* the first `produceSome()` (so no splicing), it enqueues what fits, and `pump()` is the only drainer. The 20 ms deadline is gone from the keying path entirely.
5. **Session-scoped messages** — new `M32Target`/`M32TargetScope` (RAII) narrows the tee to one transport for: both handshake replies, both goodbyes, the BLE `LINE TOO LONG`/`RX OVERFLOW` errors, and the BLE suspend/stop notices. `protocol/on`, `off`, *and* bogus values are all intercepted per transport in `bleSerialEvent`, so `m32Put`'s protocol branch is genuinely USB-only. Documented in `M32 Protocol.md` ("a goodbye you receive always means *your* session ended").
6. **`sessionResetPending`** — clear-first, like `takeLineReset()`.
7. **Downgrade hazard** — out-of-range `posBluetoothOut` now maps to 0 (Off) in `readPreferences` instead of clamping to Generic Kbd.

### Non-blocking items

- Selector value renamed to **"BLE Serial"**, and `displayValueLine` got the wrap guard so any future overlong value degrades to plain truncation instead of ghosting.
- The §10.1 splash is in: "BLE Ser. susp." is shown once when WiFi suspends BLE Serial (gated on it actually having been running, so the defense-in-depth second call doesn't double-splash). Manuals EN+DE mention the notice.
- USB `PUT device/protocol/OFF` now works — the value compare is case-insensitive on both transports (protocol description updated). I left the full shared line-assembler as future work since the intercepts above already remove the semantic drift; happy to do it as a follow-up if you'd like.
- Static `BLE2902`, scan response derived from `BLE_NAME` (+ `static_assert` on the 31-byte budget) — both as you suggested.
- JSON now reaches the tee in ~256-byte chunks via a new `MorseJSON::jsonSend()` (all 25 `serializeJson(doc, m32out)` sites converted), so a multi-KB GET is a handful of ring operations instead of thousands of per-byte call chains. This also pairs well with the new drop-don't-wait `txEnqueue`.
- **Rings stay statically allocated — deliberately.** Heap-allocating ~5.6 KB in `init()` would make BLE Serial's startup vulnerable to exactly the fragmented-heap failure mode `init()` already has to defend against (the "NO MEM" path), and it would fail at the worst time: after WiFi/BT churn. RAM headroom on both variants absorbs the static cost (V2 31.3 %, pocketwroom 39.2 %). Rationale recorded in DESIGN.md; can revisit if a variant gets tight.

### Merge gate

Agreed and unchanged: the remaining hardware checklist (WiFi suspend/resume heap deltas ×5, the classic-V2 suite incl. WiFi-upload-after-BT with free-heap numbers, snapshot/factory-reset interplay, ≥35 WpM stalled-client soak) is the gate. DESIGN.md now records your master fixes (06a3502/c61ac37) and marks the S3 `mem_release` question as the remaining open item from that thread. I'll run the Pocket side of the checklist next and post the numbers; a bench run of the V2 suite on your side would be very welcome, as offered.

vy 73, Andrew

https://claude.ai/code/session_01V541ARfCWmk6718xPh21ti

oe1wkl:
@morria — a heads-up on release planning: the current master is the release
candidate for V8.2, and this PR is too substantial for a point release. I've
therefore changed the PR's base branch to **V9.0**, the next major release
line — BLE Serial will ship with V9 rather than an 8.2.x release. V9.0 now
contains everything from master (including the BT-keyboard fixes you already
merged), so nothing from your branch is lost.

One action on your side: please merge `V9.0` into `ble-serial-protocol` to
clear the three small conflicts GitHub now shows. All three are collisions
with the V9 branch's audio-accessibility code sitting on adjacent lines, and
all resolve by keeping both sides: the include blocks + top-menu hooks in
`MorseMenu.cpp`, the include block + Bluetooth option-string array in
`MorsePreferences.cpp` (keep your renamed strings and option 5, keep the
MorseVoice lines), and `.gitignore`.

Practical upside: the remaining hardware checklist now gates the V9 timeline
instead of a release that's about to go out, so there's no time pressure.
And your static_asserts will guard the V9 preference additions too.

morria:
Hi Willi,

V9.0 as the base works well for us — agreed that this is major-release material, and no complaints about losing the time pressure. `V9.0` is now merged into the branch (`04d19ec`); the three conflicts resolved keep-both exactly as you described, with two notes:

- In `menuDisplay` the V9 side still had the bare `m32protocol` gate — kept the branch's `protocolActive()` there (rule 9), with your `MorseVoice::announce` line above it.
- **One deliberate deviation:** the `spokenName` for the *Bluetooth Use* entry now says **"Bluetooth use"** rather than V9's "Bluetooth keyboard output" — with option 5 the selector isn't keyboard-only anymore, and a blind operator navigating to it should hear what it actually covers. Trivial to change back if you prefer the original wording.

Builds are clean for the canonical pair **and** for `pocketwroom-accessibility` (so the merged `CONFIG_AUDIO_A11Y` paths — `MorseVoice::tick()` in the menu wait loop, the announce block in `displayValueLine` — actually compile, not just sit behind an off flag).

### Hardware testing since my last comment (M32 Pocket, on the bench here)

I've started the Pocket side of the checklist with the round-2 build flashed, and it has already earned its keep — **two fixes are on the branch (`9c2bcf0`) that came straight out of it:**

1. **The static `BLE2902` you suggested (and I happily adopted) doesn't survive re-init — reverted.** `BLEDescriptor::executeCreate()` bails when the descriptor's `m_handle` is already set (and `setHandle` is private, so there's no reset path). From the second `init()` cycle on, the TX characteristic silently had **no CCCD**, and every client's enable-notify failed with ATT "attribute not found" — first reproduced at the first selector off→on cycle, would equally hit every WiFi suspend/resume. The per-init `new BLE2902()` is load-bearing; it's now commented as such (the leak is small and in line with the library's other per-cycle objects).
2. **The "BLE TX backoff" DEBUG used to fire from inside `txEnqueue`** — i.e. between the chunks of a JSON object streaming through the tee — and spliced raw text into the middle of a concurrent **USB** client's object (my harness caught a corrupted `GET menus` reply during a BLE TX flood). It's now deferred and printed from `pump()`, which only runs between protocol lines.

Results so far (scripted, USB + BLE clients attached simultaneously):

- The full PLAN §8 suite (steps 1–8) **passes** on the round-2 build.
- Session isolation (fix 5) verified in **both directions**: handshake replies, goodbyes, the bogus-value error, and the suspend/stop notices each stay on their own transport; the other session stays alive and sees nothing.
- Fix 2 verified end-to-end: `PUT snapshot/recall` of a selector-0 snapshot stops BLE Serial ("BLE serial off" arrives on the BLE side, link drops, selector reads 0 afterwards); selector back to 5 restarts it via the top-menu backstop.
- Fix d verified: `PUT device/protocol/OFF` over USB ends the session properly.
- Fixes 3/4: during an 8× `GET configs` BLE flood into a full TX ring, **USB round-trip latency stayed at 42 ms** — no loop stall; drop-don't-stall behaves as documented.

One observation for the protocol description rather than the code: when the device drops a response tail under backoff, the BLE byte stream is left with an unbalanced `{` prefix, so a client that frames purely by brace-counting can't resync from the stream alone — it needs a timeout-based resync or a reconnect (fresh session reset). My test client now does the former; the iOS app will get the same. Happy to add a sentence to *M32 Protocol.md* if you agree it's worth documenting.

Still open on the Pocket checklist: the WiFi suspend/resume ×5 soak and a final full re-run on the merged V9.0 build (the bench device went to auto-sleep between runs — the sleep-suppression clearing works, at least!), plus the heap-delta measurements, which need a temporary device-side heap printout since the protocol doesn't expose free heap — I'll instrument, measure, and report the numbers here. The classic-V2 suite remains yours whenever it suits.

vy 73, Andrew

https://claude.ai/code/session_01V541ARfCWmk6718xPh21ti

morria:
Pocket hardware run complete on the merged V9.0 build (`4958227`). Everything I can exercise without a paddle or a second board now passes.

### Results (M32 Pocket, USB + BLE clients attached simultaneously)

- **Scripted PLAN §8 suite (steps 1–8): PASS** — on the round-2 build and again on the final merged build.
- **27-check dual-transport regression: 26/27.** All fix-5 isolation checks in both directions, fix 2 (snapshot store → recall → clear round-trip with selector transitions, BLE stopped/restarted correctly each time, selector value verified after recall), fix d, and the congestion policy: an 8× `GET configs` flood into a full TX ring left **USB round-trip latency at ~40 ms** (no loop stall), the backoff DEBUG arrived cleanly on raw serial, and BLE was usable again as soon as the ring drained. The single miss: the BLE suspend notice arrived in 4/5 WiFi cycles — best-effort by design (it races the link teardown), so I'm counting that as documented behavior, not a defect.
- **WiFi suspend/resume ×5: PASS** — `Disp MAC Addr` executed remotely five times; all five cycles suspended, re-advertised, reconnected and re-handshaked.

### Heap deltas (checklist item 10) — real numbers

Temporary `DEBUG` heap markers at BLE init/stop (instrumentation not committed), five suspend/resume cycles:

- Per-cycle loss, init-to-init: **3896 / 3824 / 3848 / 3824 / 3816 bytes — call it ~3.8 KB per cycle.**
- Free heap at BLE-init fell 77.6 KB → 58.4 KB across the five cycles.

That's the library's leaked server/service/characteristic/descriptor objects per `deinit(false)` → `init()` cycle, and it's noticeably more than the "few hundred bytes to ~1 KB" I had estimated in DESIGN.md. Practical reading: ~15–20 WiFi round-trips in a single power-on before `init()` could start refusing — and it refuses *visibly* through the "NO MEM" defensive path rather than crashing. I think that's acceptable for real sessions (the numbers are now recorded in DESIGN.md), but flagging it clearly since it's your merge-gate item; a real fix would mean reusing the server object across cycles, i.e. library surgery.

### One more pre-existing bug found and fixed (`4958227`)

**`GET menus` silently truncated at ~34 entries on TFT builds.** `jsonMenuList`'s fixed 3072-byte pool ran out once the games/QSO-bot menus grew `menuN` past 40, and ArduinoJson drops additions on a full pool — the entire *WiFi Functions* block was missing from the protocol menu list. Pre-existing on master for every game-enabled build; it surfaced because my suspend/resume test couldn't find `Disp MAC Addr` remotely. The pool is now computed from `menuN` (so menu growth can't re-introduce it) — verified on hardware, the full 52-entry list arrives. Worth cherry-picking to the V8.2 line if you consider it release-relevant; it's isolated to that one allocation.

### State of the merge gate

Done on the Pocket: scripted suite, suspend/resume ×5 **with** heap deltas, snapshot interplay, congestion/no-stall, session isolation, auto-sleep with an un-handshaken BLE link (confirmed incidentally — the bench device dozed off between sessions, as it should). Still open: the **classic V2** side (your bench, incl. WiFi-upload-after-BT with heap numbers), mid-game multiplayer drop, and the ≥35 WpM stalled-client soak by ear — the 40 ms loop latency is the proxy, not the proof, so I'd like a paddle session on that one before calling it closed.

vy 73, Andrew

https://claude.ai/code/session_01V541ARfCWmk6718xPh21ti

oe1wkl:
@morria — excellent bench work, thank you. Three things:

1. **The GET menus truncation is cherry-picked to the V8.2 line** (85fe8a7 on
   master, your commit, code verbatim). It's even more release-relevant than
   you thought: the QSO Bot is compiled into both variants, so the classic
   M32 (menuN = 47) was truncating too, not just the TFT builds.
2. The ~3.8 KB/cycle heap number is noted — acceptable for V9 given the
   visible NO-MEM refusal path, and now recorded. No library surgery for now.
3. Yes to the brace-resync sentence in M32 Protocol.md, please — client
   authors should learn that framing rule from the doc, not from a stuck
   parser.

The classic V2 suite is next on my bench, including WiFi-upload-after-BT
with heap numbers. On the ≥35 WpM soak I have to be honest about my own
paddle ceiling (~22 WpM), so that item moves to the V9.0 beta stage with a
high-speed operator on the key. I consider the practical risk low: while
actually keying, an operator doesn't touch parameters or menus — not even
speed — so no JSON replies are being generated mid-keying; the only BLE
traffic is the per-character echo, which is drop-on-full by design. The soak
will still be run before release, just by a faster fist than mine.

morria:
Done in `78bafcb`: the brace-resync rule is now in *M32 Protocol.md*'s Congestion bullet (timeout-resync, with reconnect as the alternative recovery), and I've corrected the `jsonMenuList` comment to note the truncation hit both variants — thanks for cherry-picking it to V8.2, and good catch on the classic `menuN` = 47.

Agreed on all three points; nothing else open on my side. Enjoy the V2 bench run — the regression harness from the Pocket pass is in `devdocs/ble-serial/round2_regression.py` (`pip install bleak pyserial`, `M32_PORT=<port>`, device at the top menu with *Bluetooth Use* = *BLE Serial*) if it saves you some typing, and the ≥35 WpM soak parked at V9.0 beta with a faster fist sounds exactly right.

vy 73, Andrew

https://claude.ai/code/session_01V541ARfCWmk6718xPh21ti

oe1wkl:
@morria — V8.2 is out, so master is now the V9 development line (the Audio
Accessibility Edition is merged in, version 9.0.0-beta). I've **retargeted this
PR to `master`** and retired the `V9.0` branch — master is the single trunk now.

Heads-up: the merge you'll need is bigger than the V9.0 one was. Master now
also carries the V8.2 snapshot-as-single-NVS-blob rework, Koch Custom Chars
(#195), and the Memory Chain game. When you merge master into
`ble-serial-protocol`, two kinds of conflict:

- **Mechanical:** wherever master still has `m32protocol` and your branch has
  `protocolActive()`, keep your side (the tee gate).
- **Needs care:** the snapshot-emitting code (`MorseJSON.cpp` configs loop and
  `MorsePreferences.cpp`) — master replaced the old `snapshotExempt()` +
  `prefName[]` reads with a new `storedInSnapshot()` / `vals[]` blob mechanism.
  Keep **master's** storage mechanism and re-apply your protocol gating on top;
  also preserve #195's empty-custom-chars-reupload guard. The manual PDFs
  conflict too — take either, the V9 manuals will replace them.

No rush given the hardware checklist is still the gate. Shout if you'd like me
to sketch the snapshot-site resolution.

oe1wkl:
Merging into master. With V9 now the beta release candidate, the remaining hardware-checklist items — classic V2 bench, the ≥35 WpM stalled-client CW-timing soak, WiFi suspend/resume heap deltas, and snapshot + factory-reset with the Bluetooth Use setting — will be validated during V9 beta rather than gating the merge.

Code review is clean: the transport tee, the per-transport gating sweep (no protocol output leaks to raw `Serial`), the single-selector mutual exclusion, and the accessibility voicing all check out.

Thank you, Andrew, for the exceptionally thorough design, implementation, and on-hardware work — and for the companion CW Keyer iOS app.

PR #195, opened 2026-07-13
Koch Custom Chars: use Koch Lesson as an index into the custom set

ckonecny:
## Summary

Currently, when the "Koch Sequence" preference is set to **Custom Chars**, the entire uploaded custom character string is always used as a flat, unfiltered pool for the CW Generator / Echo Trainer, and the "Koch Lesson" number has no effect (as documented in the user manual).

This PR makes **Koch Lesson** meaningful for Custom Chars too, mirroring how it already works for the built-in sequences (M32/LCWO/CW Academy/LICW):

- At lesson **N**, only the **first N characters** of the custom set form the active training pool, capped at the set's actual length if N is larger.
- The lesson **maximum now derives from the custom set's length** (capped to the fixed 51-slot Koch probability table) instead of being hardcoded to 51, so it adapts automatically whenever the custom set changes.

### Implementation notes

All of the bookkeeping (recomputing lesson min/max and re-clamping the current lesson) is centralized in `Koch::setCustomChars()` in `MorsePreferences.cpp`. Every code path that establishes the custom character set — menu selection of "Custom Chars", file re-upload on preferences-menu exit, the `PUT customchars/set/...` serial-protocol command, and boot-time preference loading — already funnels through `Koch::setup()` → `Koch::setCustomChars()`, so this one change propagates correctly everywhere without touching those call sites.

`getCharSet()` and `getRandomChar()` were updated to slice by the (now bounded) lesson index instead of always returning/sampling the full custom set. `getNewChar()` and `getKochChar()` were also made custom-set-aware — they previously always read from the built-in Koch table even when Custom Chars was active, which was harmless while the lesson number had no effect, but becomes a real inconsistency now that it does (they're used for "Koch: Learn New Chr" and the `GET kochlesson` protocol response, respectively).

Word/abbreviation Koch-filtering for Custom Chars (`wordIsKoch`/`createWords`/`createAbbr`) and the behavior of all other "Koch Sequence" options are unchanged.

### Docs

- Updated the English and German user manuals (`Documentation/User Manual/Version 8.x/manual_en.md`, `manual_de.md`) to describe the new Custom Chars + Koch Lesson behavior.
- Updated the serial protocol description (`Documentation/Protocol Description/M32 Protocol.md`) to note that `GET kochlesson`'s `maximum`/`characters` now reflect the custom set when Custom Chars is active.
- HTML/PDF manual regeneration (pandoc + weasyprint) was **not** run — that toolchain wasn't available in the environment this was prepared in. Flagging so the maintainer can regenerate before release if desired.

## Test plan

- [x] `pio run -e pocketwroom` — builds cleanly (M32 Pocket)
- [x] `pio run -e heltec_wifi_lora_32_V2` — builds cleanly (Classic M32)
- [x] Flashed to a physical M32 Pocket and smoke-tested: selected Custom Chars with an uploaded character set, confirmed the Koch Lesson maximum tracks the set's length, and that CW Generator/Echo Trainer only draw from the first N characters at lesson N
- [ ] Regenerate HTML/PDF manuals (left as a TODO, see above)

### DIFF
--- a/Software/src/Version 6 and newer/MorsePreferences.cpp
+++ b/Software/src/Version 6 and newer/MorsePreferences.cpp
@@ -748,8 +748,13 @@ boolean MorsePreferences::setupPreferences(uint8_t atMenu) {
                       break;
           case -1:    //////// long press indicates we are done with setting preferences - check if we need to store some of the preferences
 
-          exitFromHere: if (MorsePreferences::useCustomChars)
-                            koch.setCustomChars(getCustomChars()); //// get custom characters
+          exitFromHere: if (MorsePreferences::useCustomChars) {
+                            String chars = getCustomChars(); //// get custom characters
+                            if (chars.length() > 0)
+                                koch.setCustomChars(chars);
+                            // else: an empty re-upload keeps the previously active custom
+                            // set/bounds rather than silently wiping a working configuration
+                        }
                         if (m32protocol && posPtr < posKochFilter)
                             MorseJSON::jsonActivate(ACT_EXIT);
 
@@ -1245,8 +1250,12 @@ boolean MorsePreferences::adjustKeyerPreference(prefPos pos) {        /// rotati
                       temp = val + maxi - 2*mini + vstep  + t*vstep;
                       pliste[pos].value = (temp % (maxi - mini +vstep)) + mini;
                   }
-                  if (pos == posKochSeq)
-                      MorsePreferences::handleKochSequence();
+                  if (pos == posKochSeq) {
+                      if (!MorsePreferences::handleKochSequence()) {
+                          MorseOutput::printOnScroll(2, BOLD, 0, "No custom set");
+                          delay(700);
+                      }
+                  }
                   else if (pos == posCarouselStart && pliste[posKochSeq].value == 3)
                       MorsePreferences::handleCarouselChange();
 #ifdef CONFIG_SOUND_I2S 
@@ -1328,28 +1337,23 @@ boolean MorsePreferences::adjustKeyerPreference(prefPos pos) {        /// rotati
 }   // end of function adjustKeyerPreference
 
 
-void MorsePreferences::handleKochSequence() {
+boolean MorsePreferences::handleKochSequence() {  // returns false if Custom Chars was requested but no character set is available (yet)
     MorsePreferences::useCustomChars = false;
-    switch (MorsePreferences::pliste[posKochSeq].value) {
-      case 3:                                                 // LICW
-              handleCarouselChange();
-              break;
-      case 4:                                                 // Custom Chars
-              MorsePreferences::useCustomChars = true;
-      default:                                                // all others are treated the same
-              MorsePreferences::kochCharsLength = MorsePreferences::kochMaximum = 51;
-              MorsePreferences::kochMinimum = 1;
-              MorsePreferences::kochFilter = constrain(MorsePreferences::kochFilter, MorsePreferences::kochMinimum, MorsePreferences::kochMaximum);
-              koch.setup();
+    if (MorsePreferences::pliste[posKochSeq].value == 4) {      // Custom Chars
+        String chars = MorsePreferences::customCharSet;
+        if (chars.length() == 0)
+            chars = getCustomChars();                            // try to (re-)read it from the file player right away
+        if (chars.length() == 0)
+            return false;                                        // none available - leave the current mode/bounds untouched
+        MorsePreferences::useCustomChars = true;
+        MorsePreferences::customCharSet = chars;
     }
+    koch.setup();                                                 // setKochChars()/setCustomChars() derive the correct bounds
+    return true;
 }
 
 void MorsePreferences::handleCarouselChange() {
-      MorsePreferences::kochCharsLength = MorsePreferences::kochMaximum = koch.setupLICWkochChars(MorsePreferences::pliste[posCarouselStart].value);
-      MorsePreferences::kochMinimum = kochCharsLength > 18 ? 19 : 1;
-//DEBUG("@ 842: kMin: " + String(MorsePreferences::kochMinimum) + " kMax: " + String(MorsePreferences::kochMaximum));
-      MorsePreferences::kochFilter = constrain(MorsePreferences::kochFilter, MorsePreferences::kochMinimum, MorsePreferences::kochMaximum);
-      koch.setup();
+      koch.setup();                                               // setKochChars() case 3 derives the LICW bounds from posCarouselStart
 }
 
 
@@ -2309,17 +2313,29 @@ void Koch::setup() {                                                // create th
 }
 
 void Koch::setKochChars(uint8_t sequence) { // define the demanded Koch character set: Koch sequenze: 0 = native/JLMC, 1 = LCWO, 2 = CW Academy, 3 = LICW, 4 = Custom
+    // Bounds bookkeeping lives here (mirroring setCustomChars()) so every caller of
+    // koch.setup() gets correct kochCharsLength/kochMinimum/kochMaximum for the
+    // selected sequence, not just the ones that also call handleKochSequence()/
+    // handleCarouselChange().
     switch (sequence) {
         case 1: kochCharSet = lcwoKochChars;
                 break;
         case 2: kochCharSet = cwacKochChars;
                 break;
-        case 3: setupLICWkochChars(MorsePreferences::pliste[posCarouselStart].value);
+        case 3: {
+                uint8_t l = setupLICWkochChars(MorsePreferences::pliste[posCarouselStart].value);
                 kochCharSet = licwKochChars;
-                break;
+                MorsePreferences::kochCharsLength = MorsePreferences::kochMaximum = l;
+                MorsePreferences::kochMinimum = l > 18 ? 19 : 1;
+                MorsePreferences::kochFilter = constrain(MorsePreferences::kochFilter, MorsePreferences::kochMinimum, MorsePreferences::kochMaximum);
+                return;
+                }
         default:kochCharSet = morserinoKochChars;
                 break;
     }
+    MorsePreferences::kochCharsLength = MorsePreferences::kochMaximum = sizeof(adaptiveProbabilities) / sizeof(adaptiveProbabilities[0]);
+    MorsePreferences::kochMinimum = 1;
+    MorsePreferences::kochFilter = constrain(MorsePreferences::kochFilter, MorsePreferences::kochMinimum, MorsePreferences::kochMaximum);
 }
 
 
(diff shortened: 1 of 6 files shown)


Software/src/Version 6 and newer/MorsePreferences.cpp
```diff
@@ -2359,13 +2359,23 @@ uint8_t Koch::setupLICWkochChars(uint8_t start) {  // set up the string of chara
  
 void Koch::setCustomChars(const String& chars) {          // define the custom character set
    MorsePreferences::customCharSet = chars;
+   // Koch Lesson indexes into this set for Custom Chars, so its range must track
+   // the set's actual length (capped to the adaptiveProbabilities[] capacity).
+   uint8_t len = _min((size_t) chars.length(), sizeof(adaptiveProbabilities));
+   MorsePreferences::kochCharsLength = MorsePreferences::kochMaximum = len ? len : 1;
```
oe1wkl: **Should fix:** this clamp fires with an **empty** set in real, common flows — and then `kochFilter` (the user's saved lesson) is silently constrained to 1 and persisted on the next preferences write.

How it's reached: the preference value list wraps and `handleKochSequence()` runs on every encoder detent, so users routinely scroll *across* "Custom Chars" (MorsePreferences.cpp ~1249/1251). At that moment this function receives `MorsePreferences::customCharSet` — the NVS copy, which is `""` on any device that never used Custom Chars. The default `/player.txt` doesn't help here, because `getCustomChars()` is only called at preferences **exit** (MorsePreferences.cpp ~752). The same end state is reachable via `PUT customchars/clear` + reboot (clear leaves `posKochSeq == 4`, so boot re-enters custom mode with the empty set).

The direction I'd like to see:

1. In `handleKochSequence()` case 4, populate the set **at selection time**: use the NVS copy if it's non-empty, otherwise call `getCustomChars()` right there (side benefit: first-time selection takes effect immediately, and normal scroll-over never touches SPIFFS). Only if **both** are empty, refuse — don't set `useCustomChars`, show a short "no custom chars" note, and leave the lesson bounds alone. The refusal must not fight the scroll (no bouncing the value back).
2. The preferences-exit re-read (~752) needs the same rule: if a freshly uploaded file yields an empty set, keep the previous set (or fall back to the built-in sequence with a message) instead of adopting `""`.
3. Keep a guard **here** regardless, so no caller can ever reintroduce the clamp: when `chars` is empty, keep the full built-in range (`kochCharsLength = kochMaximum = 51; kochMinimum = 1;`) and skip the `kochFilter` clamp — i.e. replace the `len ? len : 1` fallback. The `_min` slice in `getCharSet()` already keeps the pool math safe for an empty set.

Software/src/Version 6 and newer/MorsePreferences.cpp
```diff
@@ -2359,13 +2359,23 @@ uint8_t Koch::setupLICWkochChars(uint8_t start) {  // set up the string of chara
...
+   MorsePreferences::kochMinimum = 1;
+   MorsePreferences::kochFilter = constrain(MorsePreferences::kochFilter, MorsePreferences::kochMinimum, MorsePreferences::kochMaximum);
```
oe1wkl, line 2393: **Should fix:** this bookkeeping is one-directional — it runs when *entering* Custom Chars, but nothing restores the bounds when *leaving* it through the protocol. `PUT customchars/clear` (or `set` with an empty value, m32_v6.ino ~4028–4048) flips `useCustomChars = false` and then calls `koch.setup()`, whose else-branch (`setKochChars()`) doesn't touch `kochCharsLength` / `kochMinimum` / `kochMaximum`. Until the Koch Sequence preference is changed again or the device reboots: the lesson picker stays capped at the old custom length, `GET kochlesson` reports the stale `maximum`, and `PUT kochlesson` rejects valid lessons above it.

Suggested fix: make `Koch::setup()` symmetric — have the else-branch restore the sequence-appropriate bounds (51 for the built-in sequences, or the LICW carousel length that `setKochChars()` case 3 already computes and currently discards) and re-clamp, mirroring what the custom branch now does. That would also let `handleKochSequence()` / `handleCarouselChange()` shed their now-duplicated bookkeeping.

oe1wkl (changes requested):
Thanks — this is a well-scoped, well-described PR, and its central claim holds up under review: every path that *enters* Custom Chars funnels through `Koch::setup()` → `setCustomChars()`, the boot ordering is safe (`customCharSet` is loaded from NVS before `handleKochSequence()` runs), and the word/abbreviation pools really are unchanged. There is a nice subtlety behind that last point: `wordIsKoch()`'s not-in-set sentinel is `kochCharsLength + 1`, and `setup()` passes `kochCharsLength + 1` as the threshold, so the two move together and every word still passes under Custom Chars. The `getNewChar()` / `getKochChar()` custom-awareness is a genuine fix beyond the headline feature, and `GET kochlesson`'s character loop is inherently bounds-safe now that `maximum` equals the set length. Adaptive mode also becomes coherent with custom sets (previously the lesson could exceed the set length).

Two things need fixing before merge — both stem from the same design point (the new bookkeeping only runs in one direction). Details and the direction I'd like to see are in the inline comments:

1. **An empty/unpopulated set clamps the saved lesson to 1**, reachable by merely scrolling across "Custom Chars" in the preferences (the value list wraps, and `handleKochSequence()` fires on every encoder detent). This hits devices that have *never* used Custom Chars even though a default `/player.txt` is always present, because at selection time `setCustomChars()` receives the NVS copy of `customCharSet` (empty until first use) — `getCustomChars()` only reads the file at preferences exit.

2. **Leaving custom mode via the serial protocol never restores the bounds.** `PUT customchars/clear` (or `set` with an empty value) flips `useCustomChars = false` and calls `koch.setup()`, whose else-branch doesn't touch `kochCharsLength` / `kochMinimum` / `kochMaximum`. Until the Koch Sequence preference is touched again (or a reboot), the lesson picker stays capped at the old custom length, `GET kochlesson` reports a stale `maximum`, and `PUT kochlesson` validates against it.

Smaller points, take or leave:

- `PUT customchars/set` calls `writePreferences()` *before* `koch.setup()`, so the freshly clamped `kochFilter` isn't persisted until the next write (self-heals at boot). Swapping the two lines closes the window.
- The manuals say the lesson maximum "adjusts automatically to the length of your custom character set" — please add the "(up to 51)" clause there too, as in the PR description and protocol doc. For sets longer than 51 characters, the characters beyond 51 become unreachable, which is a behavior change against the previous full-set pool.
- `sizeof(adaptiveProbabilities)` as the cap relies on the array's `uint8_t` element type; `sizeof(a)/sizeof(a[0])` (or a named constant) would survive a future type change.

Everything else checked out, including the ordering (`handleKochSequence()` writes the hardcoded 51 first, then `setup()` overrides it with the custom bounds) and the games that consume `getCharSet()`. With points 1 and 2 addressed this is a clean merge — and thank you for the thorough PR description and the docs updates, that made the review much easier.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

ckonecny:
Pushed a follow-up commit addressing both required points:

1. **Empty/unpopulated set no longer clamps the lesson to 1.** `handleKochSequence()` now tries to populate the set at selection time (NVS copy, else a fresh `getCustomChars()` read), and refuses to activate Custom Chars if both are empty — leaving mode/bounds untouched. The preferences-exit re-read no longer adopts an empty result (keeps the previous set instead). `Koch::setCustomChars()` itself also guards against an empty string as a backstop, falling back to the full built-in range rather than clamping.
2. **Leaving Custom Chars now restores correct bounds.** `Koch::setKochChars()` derives `kochCharsLength`/`kochMinimum`/`kochMaximum` itself (51 for the built-in sequences, or the LICW carousel length it already computed and previously discarded) — mirroring `setCustomChars()`. `Koch::setup()` is now symmetric regardless of branch, so `handleKochSequence()`/`handleCarouselChange()` no longer duplicate this bookkeeping and are now thin wrappers around `koch.setup()`.

Also did the smaller points: swapped `koch.setup()` before `writePreferences()` in `PUT customchars/set|clear` so the clamped `kochFilter` persists immediately; switched the array-size cap to `sizeof(a)/sizeof(a[0])`; manuals now note the 51-character cap explicitly.

**Manual test results** (physical M32 Pocket, `pocketwroom` + `heltec_wifi_lora_32_V2` both build clean):

- Custom set of 39 characters (`esno0tqr5ucd9al8ix1myj7h4gvkfz3b.6/w2p?`) → Koch Lesson maximum reported exactly **39**; at lesson 3, CW Generator drew only from the first 3 characters of the set (`e`, `s`, `n`). Confirms the core PR feature is intact.
- Koch Lesson set to 30 under "M32", then Koch Sequence stepped directly to "Custom Chars" and back to "M32" via the value-list wrap (skipping LCWO/CW Academy/LICW entirely — the exact "scroll across Custom Chars" path from your review) → Koch Lesson stayed at **30**. Confirms fix 1.
- An earlier test run happened to route through LICW Carousel on the way, and Koch Lesson correctly reported the LICW-appropriate bounds (18 for carousel start 0) purely from `koch.setup()` → `setKochChars()`, with no separate `handleCarouselChange()` bookkeeping involved. Incidentally confirms fix 2's bounds symmetry.

Separately (not part of this PR): testing surfaced that a custom set established via `PUT customchars/set` gets silently overwritten by the file player's `/player.txt` content the next time the preferences menu is exited while Custom Chars is active — that's pre-existing stock behavior (the unconditional re-read in the exit handler predates this PR), not a regression from either round of changes here. Flagging in case it's worth its own issue.

oe1wkl (approved):
Thanks for the fast and thorough turnaround — this addresses both required points exactly as discussed, and the execution is nicer than the minimum I asked for:

- The empty-set handling matches the agreed design at all three layers (selection-time populate with the NVS fast path, the preferences-exit re-read keeping a working set, and the backstop guard in `setCustomChars()`). I also verified the new boot-time implication: SPIFFS is mounted before `readPreferences()` can reach the new `getCustomChars()` path, and even a boot-time refusal falls back sanely through `setKochChars()`'s default branch.
- Moving the bounds bookkeeping into `setKochChars()` (including the LICW carousel length it previously discarded) makes `koch.setup()` genuinely symmetric and turns the two handlers into thin wrappers — and as a quiet bonus, the other pre-existing `koch.setup()` callers (snapshot recall, protocol config writes) now derive correct bounds where they could previously go stale.
- All three smaller points done (setup-before-persist in both PUT paths, the array-size idiom, the 51-cap clause in both manuals).

Verified on my side: both build environments (`pocketwroom`, `heltec_wifi_lora_32_V2`) compile clean on the PR head, and your hardware test plan covers the exact failure scenarios from the first review round.

Two cosmetic leftovers I'm noting for the record, no action needed: `PUT config` selecting Custom Chars with no set available refuses silently but still answers OK (the ignored return at m32_v6.ino ~4385 — a subsequent GET reveals the state), and the selection-time `getCustomChars()` resets the file player's word pointer — the same side effect the pre-existing exit path has always had, and only on the rare empty-NVS path.

Regarding the pre-existing issue you flagged (a `PUT customchars/set` charset being replaced by the `/player.txt`-derived set at the next preferences exit): agreed that it's stock behavior and out of scope here. We'll treat the player file as the source of truth and document the protocol PUT as session-scoped — I'll add a note to the Protocol Description on master right after this merge. Thanks for flagging it.

Merging — thanks again for a model contribution.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

ckonecny:
thanks a lot for the lightning fast progress on your end!
73 from vienna 1210!
OE1CKO :-)

oe1wkl:
Gut dass mein Freund Claude auch im Urlaub arbeitet! Schöne Grüße aus Rab, Kroatien! 73WilliVon meinem Mobiltelefon gesendetAm 14.07.2026 um 20:03 schrieb ckonecny ***@***.***>:﻿ckonecny left a comment (oe1wkl/Morserino-32#195)
thanks a lot for the lightning fast progress on your end!
73 from vienna 1210!
OE1CKO :-)

PR #196, opened 2026-07-17
Configurable CW sidetone envelope softness (M32 Pocket)

ckonecny:
## Summary
- Exposes the I2S sidetone's attack/release envelope (already used for CW keying via `cw-i2s-sidetone`'s Blackman-Harris shaping, previously hard-coded to 5 ms) as a new "Tone Softness" preference: 1-10 ms in 1 ms steps, default 5 ms (matches the prior hard-coded behavior, so existing users see no change unless they opt in).
- New preference sits right after "Tone Pitch" in every on-device menu and in the `m32_config_tool.html` WebUI's "General" group; only compiled in for `CONFIG_SOUND_I2S` builds (M32 Pocket), since Classic M32's plain PWM tone path has no such envelope.
- Changing the value applies immediately, both from the on-device encoder menu and from the serial/WebUI protocol (`PUT config/tone softness/<value>`), matching the existing `Theme` preference's apply-after-`readPreferences()` pattern.
- Updates the EN/DE user manuals with a new table row documenting the setting.

## Why
The CW sidetone can sound harsh on short elements (e.g. the single dit for "e"), especially compared to how a real transceiver rounds off keying edges with an RC low-pass filter. The underlying envelope shaping already existed in the I2S sidetone library but wasn't user-adjustable. This was validated interactively on a physical M32 Pocket (5 ms is the old default/harsher end, 10 ms felt like too much, 7 ms was a personal sweet spot) — but the shipped default stays at 5 ms so nobody's sidetone changes on upgrade without them choosing a new value.

## Test plan
- [x] Builds cleanly for `pocketwroom` (M32 Pocket) and `heltec_wifi_lora_32_V2` (Classic M32)
- [x] Flashed to a physical M32 Pocket; verified the new "Tone Softness" preference appears after "Tone Pitch", persists across reboot, and audibly softens the tone
- [x] Verified immediate apply via the on-device menu
- [ ] Not yet built for the other `CONFIG_SOUND_I2S` environments (`heltec_wifi_kit_32_V3`, `pocketwroom-lora`, `pocketwroom-170x240`, `ESP32_S3_Devkit`, `heltec-vision-master-t190`)
- [ ] Generated `manual_en.html`/`manual_de.html` and PDFs are not regenerated (pandoc/weasyprint unavailable in this environment) — markdown sources are up to date


oe1wkl:
This has to wait a few days . I am currently hunting a NVS storage problem, and that has to be resolve before we add more NVS items…
Willi

oe1wkl:
Thanks Christian — nice feature, and the sidetone really does benefit from this. I've reviewed it against current master; a few things to address before I merge.

**Verified up front:** it still merges cleanly on current master (which has moved on quite a bit — snapshot blob rework, NVS budgeting, V9.0 merge), and builds clean for `pocketwroom`, `heltec_wifi_lora_32_V2` and `pocketwroom-accessibility`. The `prefPos` / `pliste[]` / `prefName[]` three-array sync is done correctly, `PREFPOS_COMMON_CORE` reaches all 13 option lists, the NVS key is 13 chars (under the 15 limit), and the envelope only wraps the sine generator — so the V9.0 accessibility voice clips are unaffected. All good.

I also checked the timing implications, since anything touching the keying envelope makes me nervous. Good news: **external transmitter keying is not affected at all.** `keyOut()` asserts `keyTransmitter()` and `digitalWrite(keyerPin, LOW)` *after* the blocking delay on both edges, so both shift equally and cancel — the keyed mark stays exactly the scheduled element duration, and `setADSR()` only touches the audio chain. And the Blackman-Harris envelope itself is duration-neutral: because `build_fall()` is `1 - bh(i)`, both edges cross −6 dB at the same fraction of the ramp (φ = 0.6568), which makes it a pure delay rather than something that stretches elements. So no timing correction is needed for this preference. There *is* a separate, pre-existing question about how the sidetone lines up with the key line, but that one predates your PR and we'll take care of it ourselves — see the note at the end.

---

## 1. Blocker: `setADSR()` races the audio task

`setADSR()` currently has **no other call site anywhere in the firmware** — the envelope is built once in the `I2S_Sidetone` constructor and never touched again. This PR introduces the first runtime call, which exposes something that was previously unreachable.

`BlackmanHarrisEnvelope::build_rise()` / `build_fall()` call `std::vector::resize()`, reallocating the exact buffers the audio task indexes per-sample in `process()` — and that task is pinned to core 1 at `configMAX_PRIORITIES - 1`. Worst case is reached by design here: `setSidetoneEnvelope()` is called from `adjustKeyerPreference()`, so it fires on **every encoder detent**, while the encoder click sound is playing through that same envelope. Each rebuild is also ~900 double-precision `cos()` calls (software FP on ESP32), which widens the window considerably.

Suggested fix: give the envelope fixed-capacity arrays plus an integer length, and have `setAttackRate`/`setReleaseRate` only write that length. At 10 ms / 44.1 kHz that's 441 floats per table, ~3.5 KB total — affordable, and nothing ever reallocates. The existing bounds checks in `process()` already handle a length that shrinks under a live cursor.

## 2. Please cap the range at 9 ms

Partly your own finding — you reported 10 ms as too much and 7 ms as the sweet spot, so nothing useful is lost. But mainly: the follow-up work mentioned below needs the value to stay at or under 9 ms, and shrinking a user-visible range after it has shipped (with values already in NVS and in snapshots) is a lot more awkward than starting narrow. Easier to do now.

## 3. Snapshots — master moved under you

`storedInSnapshot()` returns `true` by default, so `posSidetoneShape` currently lands in snapshots. Master's `dfe1786` rescoped snapshots to *training settings only*, and `posLineOut` — the closest sibling — is explicitly excluded. Please add `posSidetoneShape` to that exclusion list. It also has no apply-hook on snapshot recall, so a recalled value would sit in `pliste` unapplied; excluding it fixes both at once.

## 4. Small stuff

- `MorseOutput.h` says the parameter "indexes `sidetoneEnvelopeMs[]`" — no such array exists.
- `MorseOutput.cpp` says "prefValue 0..14 maps to 1..15 ms"; the pref is 0..9 → 1..10 ms. Both look like leftovers from an earlier draft.
- Minor correction to the PR description: `Theme` is applied in `writePreferences()` (`MorsePreferences.cpp:1787`), not after `readPreferences()` — which is why Theme survives a snapshot recall and this wouldn't.

---

Items 1–4 and it's good to go from my side. Don't regenerate the manual HTML/PDFs — I'll do that in the release build.

One last note so the 9 ms cap doesn't look arbitrary: while reviewing this I found that our I2S sidetone sits about 6 ms "heavier" than the key line, dating back to #136/#138 — the fix there matched the two audio paths' blocking *time* but not their acoustics (on PWM those 6 ms fade the tone, on I2S they hold it at full amplitude). That's our legacy issue, not yours, and I've written it up in `devdocs/cw-timing-audit/FINDINGS.md` as a next step for us. The planned fix redistributes the existing 6 ms as `delay(6-k) / on() / delay(k)` with `k = round(0.657 · softness)`, which is where the ≤ 9 ms constraint comes from. Your preference is what makes that fix worth doing, so thanks for prompting it.

ckonecny:
Thanks for the thorough review — addressed 1-4:

1. **Race fix**: `BlackmanHarrisEnvelope` now uses a fixed-capacity `std::array` (sized for ~20ms @ 48kHz) plus a published length instead of `std::vector::resize()`. `build_rise()`/`build_fall()` only overwrite the table in place and publish the new length last, so the audio task reading it per-sample on the other core never sees a reallocated/torn buffer.
2. **Range capped to 1-9ms** (was 1-10ms), leaving headroom for the timing fix in `devdocs/cw-timing-audit/FINDINGS.md`.
3. **Snapshot exclusion**: added `posSidetoneShape` alongside `posLineOut` in `storedInSnapshot()`.
4. **Stale comments fixed** (dead `sidetoneEnvelopeMs[]` reference, wrong `0..14` range). Noted on the Theme correction too — thanks, description was imprecise there, no code change needed since this pref is now snapshot-excluded (no recall-apply gap to cover).

Also built all `CONFIG_SOUND_I2S` environments as requested. `pocketwroom`, `heltec_wifi_lora_32_V2`, `pocketwroom-accessibility`, `pocketwroom-170x240`, and `heltec-vision-master-t190` all build clean. Four others fail, but on current master *without* this branch's changes too, so pre-existing and unrelated to this PR:
- `heltec_wifi_kit_32_V3`, `heltec_wifi_lora_32_V3`, `ESP32_S3_Devkit`, `minipcb_lora`: `MorseOutput::soundEventHandler()` is called under `CONFIG_SOUND_I2S` in `adjustKeyerPreference()` but only declared under the narrower `CONFIG_TLV320AIC3100` — breaks every WM8960 board.
- `pocketwroom-lora`: missing `RadioLib.h` — that env's `lib_deps` doesn't pull in RadioLib.

Happy to open separate PRs for those if useful, but didn't want to bundle unrelated fixes into this one.

oe1wkl:
Merging — thanks for the quick turnaround.

I verified all four: the fixed-capacity `std::array` closes the race (good catch extending the copy constructor to carry the new length fields), range is 1–9 ms, `posSidetoneShape` is excluded from snapshots alongside `posLineOut`, comments fixed. Built your head for `pocketwroom`, `heltec_wifi_lora_32_V2` and `pocketwroom-accessibility` — all clean.

I also reproduced both pre-existing failures on plain master: `heltec_wifi_kit_32_V3` gives `'soundEventHandler' is not a member of 'MorseOutput'` — declared under `CONFIG_TLV320AIC3100` but called under `CONFIG_SOUND_I2S`, so it breaks every WM8960 board — and `pocketwroom-lora` is missing RadioLib. You were right not to bundle them; I'll take those separately, no need for you to open PRs.

Two things that are on me rather than you:

- The Accessibility Edition needs voice clips for "Tone Softness" and the nine "N ms" option values, plus pronunciation overrides so they aren't read out as "em ess". That duty only went into `CLAUDE.md` after you opened this PR and I hadn't pushed it yet — my oversight, not something you could have known. It needs the Piper toolchain, so I'm doing it as part of the merge.
- Manual HTML/PDF regeneration, as mentioned earlier.

One non-blocking note if you ever revisit the envelope: `attack_len_` / `release_len_` are plain `int`, so "publish the length last" isn't formally ordered — the array and the scalar provably don't alias, so a compiler may legally sink the fill loop below the length store. `volatile` or `std::atomic<int>` would make the guarantee real at zero cost. Very unlikely to bite in practice, which is why I'm not holding the merge for it.

PR #197, opened 2026-07-19
Practice Stats: Koch-lesson time & per-character error tracking (M32 Pocket)

ckonecny:
## Summary

Adds opt-in practice statistics for the Koch trainer on M32 Pocket: how long you've spent at each Koch lesson (split by listening vs. sending), and your per-character error rate in the Echo Trainer.

- New `MorsePracticeStats` module logs one JSON-lines record per contiguous practice segment to `/stats.jsonl` on SPIFFS: lesson, listen/send mode, duration, words presented, per-character attempt/error counts, and the WPM/Interchar/InterWord speed settings in effect. Size-capped (128 KB) with oldest-first trimming.
- No RTC on this hardware, so timestamps come from whichever sync happens first: opportunistic NTP once WiFi connects, the browser's clock when the stats page loads, or `put time/set/<epoch>` over the serial protocol. Uses `settimeofday()` so a sync survives normal sleep/wake cycles (not the physical power switch, which cuts the RTC domain — sessions logged before the next sync just show as "(undated)" rather than losing data).
- New WiFi menu item **WiFi Functions → Practice Stats** (`MorseWiFi::viewStats()`) serves a self-contained page: time-per-lesson bars (stacked green=listen/blue=send), per-character error-rate table, full session history (words, chars/word, WPM, spacing, WPS/CPS, errors), Clear Log.
- Same data exposed over the existing serial JSON protocol (`get stats/log` — base64-encoded, since the log's raw `{`/`}` would break the existing brace-counting response parser; `put stats/clear`; `put stats/enabled`) and surfaced in the M32 Configuration Tool's Koch tab, for USB-only users.
- New **Practice Stats: On/Off** preferences item (in the full preferences list, reachable from the top-level menu) to disable logging entirely.
- Gated by `CONFIG_PRACTICE_STATS`, set only for Pocket/TFT build environments — Classic M32 compiles the feature out completely.
- EN + DE user manual sections; developer notes in `devdocs/practice-stats/README.md` covering the log format, timestamp design, and known limitations.

## Test plan

- [x] `pio run -e pocketwroom` and `pio run -e heltec_wifi_lora_32_V2` both build clean, no warnings.
- [x] Flashed to a physical M32 Pocket; verified on-device: menu navigation (both encoder directions) reaches Practice Stats; Echo Trainer and CW Generator sessions are logged with correct lesson/duration/word counts; lesson-change and sleep both flush segments correctly.
- [x] Verified the WiFi stats page end-to-end: connects via `WiFi Select`'s active network (not its own AP), browser time sync, session history renders and updates, Clear Log works, tables scroll horizontally without breaking page layout.
- [x] Verified the M32 Configuration Tool's new Koch-tab section against the same device over USB: loads/refreshes, Clear Log, Logging On/Off toggle.
- [ ] Not tested: NTP sync against a real internet-connected AP (only verified the browser-clock and serial fallbacks on the available test network).


oe1wkl:

Impressive scope, and the design notes in `devdocs/practice-stats/README.md` are excellent — the `statsDiscardWord`-vs-`stopFlag` reasoning and the `settimeofday()` / backfill design are spot-on. I also confirmed the base64 chunking is correct (48 is a multiple of 3, so only the final chunk pads) and the two `recordWord()` sites sit in mutually-exclusive `if/else` branches, so there's **no double-counting**. ✅ **Verified builds clean** on all three envs — `heltec_wifi_lora_32_V2`, `pocketwroom`, and `pocketwroom-accessibility`.

**Two things to fix before merge:**

1. **Hard-rule §9 — `MorseJSON::jsonStatsLog()` writes raw `Serial`.** Every `Serial.print(...)` / `Serial.write(...)` in that function should go through the `m32out` tee. Its sibling `jsonFileText()` right above it was explicitly converted to `m32out.print`/`m32out.write` for exactly this reason (BLE Serial transport). As written, `get stats/log` returns nothing over BLE Serial — it works over USB, which is why the config-tool test passed. Mechanical fix: `s/Serial\./m32out./` in that function.

2. **Accessibility §8 — the new "Practice Stats" UI text is mute.** The new menu item (WiFi Functions) and the preference label are compiled into `pocketwroom-accessibility` (it inherits `M32_Pocket_TFT_build_flags` and only unflags `CONFIG_CW_GAME`), but the PR adds no voice clip, so they're silent for blind operators. #199 handled the analogous case (added the "Practice Set" clip). Fix: run `extract_voice_strings.py` + `generate_audio.sh` and commit the clip. ("On"/"Off" values are already voiced.)

**Cross-PR notes (with #199):** both PRs append to the `prefPos` enum + `extraItems[]` (positional `extraItems[pos - posKochFilter]` mapping) and both edit the `menuNo`/`menuText`/`menuNav`/`menuN` block — enum-append order and array order must stay in lockstep, and `menuN` must resolve to `45 + (CONFIG_PRACTICE_STATS ? 1 : 0)`. Also, #199's Quick-Start exclusion switch needs a `case _wifi_stats:` added for this feature (see the note on #199).

**Minor (cosmetic):** the char-tile grid keys off `/api/koch` characters run through `cleanUpProSigns()` (e.g. `<sk>`), but the log's `chars` keys are the raw prosign bytes — so for a lesson containing prosigns those tiles read "heard 0." TFT-only, display-only.

We're happy to prepare the §9 and §8 fixes (and the `_wifi_stats` reconciliation) so this is merge-ready.

oe1wkl:
Update: your latest push resolves **both** items I raised — thanks, nothing further from me on §8/§9.

- **§8 (accessibility):** `d2bdb02` voices "Practice Stats" (`eaac0f5b.mp3`), so the menu item and preference label are no longer silent in the accessibility build.
- **§9 (`m32out`):** the `upstream/master` merge (`40ea2f0`) that brought in the `m32out` tee also converted `jsonStatsLog()` to `m32out.print`/`m32out.write`, so it's now consistent with `jsonFileText()` and `get stats/log` will reach BLE-serial clients too.

That also retro-corrects my earlier §9 note: on the pre-merge branch it wasn't actually an in-branch defect (`jsonFileText` was raw `Serial` there too) — it only became actionable once master's `m32out` landed, which your merge has now handled. Apologies for the earlier over-scoping.

I have **not** re-built the freshly-merged branch (the merge pulls in a lot, incl. the BLE-serial work) — happy to run the two-variant + accessibility build to confirm the merged state is clean if that's useful.

PR #199, opened 2026-07-24
Add Practice Set for CW Generator and Echo Trainer

ckonecny:
### Summary
- **What it does:** "Practice Set" is a new content type for the CW Generator and Echo Trainer, alongside Random/CW Abbrevs/English Words/etc. It lets the operator pick a handful of individual characters directly on the device — e.g. just the few they personally find difficult — using an encoder wheel where each character can be added at most once. Whichever characters were picked then become the entire pool that mode draws from when generating practice text, so training is focused purely on that self-chosen set rather than a lesson sequence or a random mix.
- Deliberately separate from the Koch Trainer's own "Custom Chars" lesson data (own prefPos, own globals, own NVS key), so neither feature can clobber the other. Unlike the original session-local design, the set now persists across power-off.
- The settings-menu picker sits in the general Settings menu right after "LICW Carousel", and is also editable from the web config tool (Preferences → Other), backed by a new `get`/`put practicechars` serial-protocol pair mirroring `customchars`.
- EN/DE manuals updated for the new feature.
- The "Practice Set" menu/preference labels are voiced for the Accessibility Edition (one new clip, `2c91ef54.mp3`).

### Test plan
- [x] Builds clean: `heltec_wifi_lora_32_V2`, `pocketwroom`, `pocketwroom-accessibility`
- [x] Flashed and tested on physical M32 Pocket hardware — Practice Set picker, CW Generator/Echo Trainer drill modes
- [x] Web config tool: Practice Set row in Preferences → Other, tested against the device over WebSerial
- [x] Accessibility Edition: flashed + `uploadfs`'d to hardware, "Practice Set" clip confirmed audible on-device
- [ ] EN/DE manual HTML/PDF not rebuilt (no `pandoc`/`weasyprint` in this environment) — markdown sources are done, `build.sh` needs a local run


oe1wkl:

This one's the cleanest of the three. ✅ **Verified builds clean** on `heltec_wifi_lora_32_V2`, `pocketwroom`, and `pocketwroom-accessibility`. I checked the conventions that usually bite here and they're all handled correctly:

- **Action-pref plumbing** (CLAUDE.md §3 rule 10): in `extraItems[]` + `getValueLine`/`adjustKeyerPreference` + `allOptions`, kept **out** of `pliste[]`/`prefName[]`, repaints via `displayKeyerPreferencesMenu(pos)`. The positional `extraItems[pos - posKochFilter]` mapping stays aligned.
- **`practiceCharSet` is fully separate** from Koch `customCharSet` — own NVS key `practiceChars` (13 chars, under the 15-char limit), reset via `usePracticeChars = false` in `menuExec()` so it can't leak into other modes.
- **Length is safely bounded:** `posRandomLength`'s max maps to "2 to 6" and `random(2, maxLength-3)` caps at 6, so the practice loop's `reserve(7)` and uncapped `for` are fine.
- **`PRACTICE_WHEEL_LEN = 51`** correctly stops before the multi-byte `äöüH` tail of `CWchars`.
- **§8 handled** — voice clip added and all three extractor outputs updated.
- The `noDuplicates` / `displayXform` extension to `MorseTextEntry` is clean and respects the prosign hard-rule (§3 #4).

**One cross-PR note** (only matters if this merges alongside #197): the "Fix Quick Start not remembering Practice Set" commit rewrites the `writeLastExecuted` test in `menu_()` into an explicit exclusion switch. That switch has no `case _wifi_stats:` — which only exists in #197. If both land, Practice Stats (a WiFi function) would wrongly become a Quick-Start / last-executed target. Whoever merges second should add `case _wifi_stats:` (under `#ifdef CONFIG_PRACTICE_STATS`). Also flagging that #197 and #199 both append to the `prefPos` enum + `extraItems[]` **and** both touch the `menuNo`/`menuText`/`menuNav`/`menuN` block, so the enum-append order and array-append order must stay in lockstep on merge (positional mapping).

Minor: `setPracticeChars()` doesn't check the `begin()`/`putString()` result (§4 rule b) — consistent with `playerCall`/`playerName`, so low priority.

No blockers from our side. We're happy to send the trivial `_wifi_stats` reconciliation as part of whichever of #197/#199 lands second.

oe1wkl:
Pushed `71bf9fb` (maintainer edit) for the Quick Start fix.

Rather than add a `case _wifi_stats:` to the exclusion list, I replaced the denylist `switch` with an allowlist:

```cpp
if (MorsePreferences::menuPtr < _wifi
    || MorsePreferences::menuPtr == _genPractice
    || MorsePreferences::menuPtr == _echoPractice)
    MorsePreferences::writeLastExecuted(MorsePreferences::newMenuPtr);
```

This restores the original `< _wifi` semantics (all training modes + the classic games) plus the two Practice Set leaves, so every WiFi function, Go To Sleep, grid game — and any *future* append-only WiFi/game leaf, including #197's `_wifi_stats` — is excluded by construction, with no per-leaf edit ever needed. Behaviour is identical to your switch for every current menu entry; this just removes the maintenance hazard, so the cross-PR reconciliation with #197 I flagged earlier is no longer needed.

Builds clean on `heltec_wifi_lora_32_V2`, `pocketwroom`, and `pocketwroom-accessibility`.

PR #200, opened 2026-07-27
Fix Custom Chars silently reset by any preferences-menu exit

ckonecny:
### Summary
- **The bug:** When the Koch Trainer's "Custom Chars" character set is active, exiting the preferences menu for *any* reason — even changing an unrelated setting like Interword Space — silently replaced the active custom character sequence with whatever text is currently uploaded to the file player, discarding whatever sequence was actually in use.
- **Root cause:** `setupPreferences()`'s exit handler unconditionally re-derives the custom set from `/player.txt` via `getCustomChars()` whenever `useCustomChars` is true, regardless of what was actually changed. This should only happen when the user explicitly re-selects "Custom Chars" as the Koch Sequence value (as documented in the manual) — which is exactly what `handleKochSequence()` already does correctly elsewhere, falling back to the file only when no custom set is active yet.
- **The fix:** Bring the preferences-exit handler in line with `handleKochSequence()`'s existing guard — only pull from the file player when `customCharSet` is currently empty (bootstrap case), never overwriting an already-active set just because the user exited preferences for an unrelated reason.
- Found and reported by Christian Konecny while testing an unrelated branch.

### Test plan
- [x] Builds clean: `heltec_wifi_lora_32_V2`, `pocketwroom`
- [x] Reproduced on physical M32 Pocket hardware: set a custom sequence, changed Interword Space, exited preferences — sequence was replaced with file-player content
- [x] Confirmed fixed with the same steps after the patch — sequence preserved

### DIFF
--- a/Documentation/Protocol Description/M32 Protocol.md
+++ b/Documentation/Protocol Description/M32 Protocol.md
@@ -546,12 +546,14 @@ Example:
 
 	PUT customchars/set/mkrsuaptlowi
 
-Note: on the device, the custom character set is derived from the uploaded
-player file (`/player.txt`) — that file is the source of truth. A set injected
-with this command therefore persists only until the preferences menu is next
-exited while "Custom Chars" is active: at that moment the set is re-derived
-from the player file, replacing the injected one. To make a character set
-permanent, upload it as the player file instead (or in addition).
+Note: on the device, the custom character set can also be derived from the
+uploaded player file (`/player.txt`). A set injected with this command persists
+across preferences-menu visits and stays active until explicitly replaced — it
+is *not* overwritten by exiting the preferences menu. It is only re-derived
+from the player file when "Koch Sequence" is explicitly (re-)selected to
+"Custom Chars" (via the on-device menu, or via `PUT config/koch sequence/4`),
+so uploading a new player file and then re-selecting "Custom Chars" is how you
+load its content, overwriting anything injected with this command.
 
 `PUT customchars/clear`
 
--- a/Software/src/Version 6 and newer/MorsePreferences.cpp
+++ b/Software/src/Version 6 and newer/MorsePreferences.cpp
@@ -1319,7 +1319,7 @@ boolean MorsePreferences::adjustKeyerPreference(prefPos pos) {        /// rotati
                       pliste[pos].value = (temp % (maxi - mini +vstep)) + mini;
                   }
                   if (pos == posKochSeq) {
-                      if (!MorsePreferences::handleKochSequence()) {
+                      if (!MorsePreferences::handleKochSequence(true)) {    // explicit on-device (re-)selection: force a file reload
                           MorseOutput::printOnScroll(2, BOLD, 0, "No custom set");
                           delay(700);
                       }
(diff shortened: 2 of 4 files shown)


oe1wkl:

The one-line fix is correct for the reported bug and correctly mirrors `handleKochSequence()`'s empty-guard. ✅ **Verified builds clean on both variants** — `heltec_wifi_lora_32_V2` and `pocketwroom`.

One thing to resolve before merge — the fix also changes behavior the manual documents:

Pre-fix, the unconditional re-read on preferences-exit did double duty: it clobbered *injected* sets (the bug you're fixing here) **but also** reloaded an *updated* `/player.txt` into an already-active set — which is the manual's documented "change your character set" procedure (§5.4.1: *"upload a new text file and select Custom Chars again … This is a feature, not a bug!"*). After this change, neither the exit handler nor `handleKochSequence()` re-reads the file while `customCharSet` is non-empty (and it's never cleared on a Koch-Sequence switch), so the documented reload workflow silently stops working. The PR premise ("re-derive only when the user explicitly re-selects Custom Chars, which `handleKochSequence()` already does") rests on behavior that doesn't actually exist — `handleKochSequence()` only re-reads when the set is empty.

Two clean ways forward:
1. **Preserve the workflow:** make `handleKochSequence()` re-read from the file on an *explicit (re-)selection* of Custom Chars (the deliberate gesture), keeping this PR's exit-handler guard. Fixes the clobber **and** keeps "re-select to reload."
2. **Keep as-is** and update the docs to describe the new "sticky set — clear, then re-set" behavior.

Either way, two docs currently describe the old behavior and need updating in this PR:
- `Documentation/Protocol Description/M32 Protocol.md` (~L582-587) now states the *opposite* of reality (it says an injected `PUT customchars/set` is dropped at the next prefs-exit — after this fix it persists).
- `manual_en.md` / `manual_de.md` §5.4.1 "change the character set" paragraph.

We're happy to prepare whichever patch is preferred (leaning toward option 1 + doc sync) so this is merge-ready.

oe1wkl:
Verified `e4780e2` — this is exactly the Option 1 approach, and it's correct:

- `handleKochSequence(forceReload)` re-reads the file player only on an **explicit (re-)selection** of Custom Chars, with both call sites covered — the on-device Koch Sequence encoder *and* the serial `PUT config/koch sequence/4` path (nice catch on the latter). Boot and snapshot-recall keep the default (`false`) and preserve the persisted set.
- The exit-handler guard from `5328f1c` still fixes the original clobber: an unrelated preference change never calls `handleKochSequence()`, and the guard skips the file re-read while a set is active.
- `M32 Protocol.md` rewritten to match; the EN/DE manuals were already accurate.

I traced the three key paths and they all behave as intended:
- inject a set → change an unrelated pref → exit: **injected set survives** ✅
- upload a new `player.txt` → re-select Custom Chars: **new content loads** ✅
- reboot: **persisted set is used** (no file clobber) ✅

Builds clean here on `heltec_wifi_lora_32_V2` and `pocketwroom`.

Agree with your follow-up note: the "Custom Chars selected but no set available" state (menu shows Custom Chars while the engine keeps the previous sequence) is pre-existing on master and worth a separate ticket — out of scope for this fix. No blockers from me on this PR.

PR #201, opened 2026-08-03
Practice Stats: show storage usage as a labeled breakdown, not a raw total

ckonecny:
## Summary

The "Storage" line on both the WiFi Practice Stats page and the config tool's Practice Stats section reported `SPIFFS.usedBytes()`/`totalBytes()` — the whole filesystem — under a section titled "Practice Stats", which reads as if it were the log's own size. A user with only ~4 KB of actual log data (16 sessions) saw "877 KB / 1405 KB used (62%)" and reasonably assumed the log itself had ballooned.

- Added `MorsePracticeStats::logBytes()` (the actual `/stats.jsonl` size) and threaded it through the `get stats/log` serial response and the WiFi page's `/api/storage` endpoint as `logSize`.
- Both UIs now show a labeled stacked bar (practice log / other / free) plus the same breakdown spelled out in the legend labels themselves (KB and % inline), replacing the previous single ambiguous line (and an intermediate table iteration before that).

## Test plan

- [x] Builds clean for `heltec_wifi_lora_32_V2` and `pocketwroom`
- [x] Verified against physical M32 Pocket hardware: read the live `used`/`total`/`logSize` numbers via the serial protocol before and after deleting leftover accessibility-edition voice clips, and again after a full chip erase + reinstall — confirmed "other" tracks real non-log storage (dropped from ~873 KB to ~0.5 KB) while the practice-log number tracked the actual `/stats.jsonl` size throughout
- [x] Visually confirmed the new storage bar + labeled legend on both the WiFi stats page and the config tool


oe1wkl:
Reviewed — this is sound, and the ambiguity it fixes was a real one. I built the branch for both variants myself (`heltec_wifi_lora_32_V2` and `pocketwroom`): clean. `logBytes()` is safe as written — the log is opened and closed per append, so `.size()` is never stale, and `File::size()` self-guards a failed open.

Two things before I merge:

1. **Dead CSS.** `.bar .log` / `.bar .other` in `MorseWiFi.cpp` and `.stats-bar .log` / `.stats-bar .other` in the config tool are added but never used — the storage bar uses `.storagebar` / `.storage-bar`. And `#storage{...}` in `MorseWiFi.cpp` is now orphaned, since its `<div>` was removed. Leftovers from the earlier table iteration, I assume. The WiFi page lives in flash, so please drop them.

2. **Manual, EN + DE.** The Practice Stats section lists what the page shows; "Storage" is now a prominent heading there and isn't mentioned. One bullet in `Documentation/User Manual/Version 9.x/manual_en.md` and the same in `manual_de.md`.

Nits, not blockers: `StaticJsonDocument<64>` in `viewStats()` is now at 48 of 64 bytes, i.e. room for exactly one more field; `logPct` goes `NaN` if `total` is 0 (unmounted SPIFFS); and "other" absorbs the log's SPIFFS page rounding, so the `.1 KB` precision reads slightly more exact than it is.

Separately, and *not* something for this PR: `GET stats/log` is entirely absent from `Documentation/Protocol Description/M32 Protocol.md` — the whole `stats` command family arrived undocumented with the Practice Stats commit. I'll pick that up as a protocol 1.4 doc item.

PR #202, opened 2026-08-06
CW Generator: wrap display at word boundaries, not mid-word

ckonecny:
## Summary

- Wraps the scroll display's line breaks at word boundaries for the CW
  Generator (`morseState == morseGenerator`), instead of breaking wherever
  the character count happens to run out — a group like "excc" no longer
  gets split as "exc" / "c" across two lines.
- Scoped exclusively to `morseState == morseGenerator` (covers both direct
  CW Generator and Koch Trainer → CW Generator, since both drive this
  state) via a small addition in `dispGeneratedChar()` (`m32_v6.ino`): it
  looks ahead in `clearText` — which already holds the rest of the current
  word — to check whether the word fits the remaining line width, and
  forces a line break beforehand if not.
- Echo Trainer, CW Keyer, decoded-audio display, Transceiver (LoRa/WiFi)
  playback, games and QSO Bot are untouched and behave exactly as before.
- `MorseOutput.cpp`: no behavior change to `printToScroll_internal()`
  itself — same wrap logic, same order, same conditions. The only change
  is that its `screenPos` counter moved from a function-local `static` to
  a translation-unit-scope `static` (`scrollScreenPos`), so a new
  read-only query `MorseOutput::wordNeedsWrap(uint16_t wordLen)` can read
  it. Both are `static` storage duration, so this is a visibility change
  only, not a lifetime change. `dispGeneratedChar()` and
  `printToScroll_internal()` both run synchronously in the main loop —
  no FreeRTOS task or ISR boundary is crossed — so this introduces no new
  race; flagging proactively given this file's history (#196).

## Why

Feedback from testing on a physical M32 Pocket (LCD): when copying CW and
writing it down in fixed-size groups (5-character groups, but the same
applies to single characters, 2/3-character groups, or free-form words),
a group would sometimes get split across the display's line wrap — one
or two characters stranded at the end of a line, the rest starting the
next. That makes comparing the display against handwritten copy harder
than it needs to be, since the on-screen "columns" no longer line up with
the actual groups.

This is only fixable for the CW Generator because it already holds the
complete upcoming word in `clearText` before revealing it
character-by-character (paced to the CW audio) — the fix just looks
ahead before the first character of a new word. Live keyed/decoded CW
doesn't have this property: the device genuinely can't know how long the
current "word" is until the operator pauses, so those paths are
intentionally left as they are.

## Open question

Should this always be on, or would you rather it be an opt-in preference
(e.g. "Word Wrap" under General settings, default **off** so existing
users see no change unless they choose it)?

Leaning towards "just fix it" — it only ever produces the same or better
line breaks than the old character-count wrap, no case where it looks
worse. But flagging the alternative since a toggle would need its own
manual entry (EN+DE) and an accessibility voice clip (§8), both of which
the always-on version avoids entirely. Happy to add the preference if
you'd rather users opt in explicitly.

## Test plan

- [x] Builds clean for `heltec_wifi_lora_32_V2` (Classic M32) and
      `pocketwroom` (M32 Pocket)
- [x] Flashed to a physical M32 Pocket; CW Generator tested across
      Callsigns, Mixed, Practice Set, Random, CW Abbrevs, and English
      Words — groups/words of varying length never split across a line
      break
- [x] Koch Trainer → CW Generator tested the same way, including lessons
      where word length varies within one session — wraps correctly
      every time
- [x] Confirmed Echo Trainer / CW Keyer / decoded-audio display is
      visually unchanged (still wraps by character count, as before)

## Scope

No new preference, NVS key, menu entry, or protocol command — nothing to
update in the manuals, Protocol Description, or `prefPos`/`extraItems`/
snapshot bookkeeping, unless the preference above is wanted, in which
case all of that would apply.

### DIFF
--- a/Software/src/Version 6 and newer/MorseOutput.cpp
+++ b/Software/src/Version 6 and newer/MorseOutput.cpp
@@ -82,6 +82,7 @@ char textBuffer[NoOfLines][2 * NoOfCharsPerLine + 1]; /// we need extra room for
 
 uint8_t linePointer = 0;    /// defines the current bottom line
 uint8_t bottomLine = 0;
+static uint8_t scrollScreenPos = 0;   /// current column on the scroll line; used by printToScroll_internal() and by wordNeedsWrap()
 
 const int8_t MorseOutput::maxPos = NoOfLines - NoOfVisibleLines;
 int8_t MorseOutput::relPos = MorseOutput::maxPos;
@@ -809,7 +810,6 @@ void MorseOutput::printToScroll_internal(FONT_ATTRIB style, const String& text, 
   //unsigned char ch;
   //
   static uint8_t pos = 0;
-  static uint8_t screenPos = 0;
   static FONT_ATTRIB lastStyle = REGULAR;
   uint8_t l = text.length();
   if (l == 0) {                               // an empty string signals we should clear the buffer
@@ -817,7 +817,7 @@ void MorseOutput::printToScroll_internal(FONT_ATTRIB style, const String& text, 
       textBuffer[i][0] = (char) 0;                    /// empty this line
     }
     refreshScrollArea((NoOfLines + bottomLine - (NoOfVisibleLines - 1)) % NoOfLines);
-    pos = screenPos = 0;                                // reset the position pointers
+    pos = scrollScreenPos = 0;                                // reset the position pointers
     return;
   }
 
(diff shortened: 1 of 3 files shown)


oe1wkl:
Merged to `master` in a00384e — thanks, Christian, good catch and the right mechanism. To your open question: applied always-on, no preference.

Three things changed on integration (2d77c0b):

1. **The look-ahead now runs only for the first character of a word, and only if the word fits a line at all.** As submitted it ran before *every* character and never asked whether the word fits, so for a word wider than the line the condition stayed true after each wrap and the word came out one character per line. That bites on the classic M32 (14 columns) with English Words at length "unlimited" — the list reaches 17 characters — and on both variants with long file-player tokens.
2. **The line break is display-only.** It went through `displayGeneratedMorse()`, which also feeds `SerialOutMorse()` and the Bluetooth keyboard, so a screen-layout decision would have pushed `"\n"` into the protocol character echo and typed a Return on a paired host.
3. `wordNeedsWrap()` sets the scroll font before measuring `"A"` — `printOnStatusLine()` leaves the 12 pt font active, so on TFT it could have counted columns in the wrong font. OLED is unaffected (it uses the constant).

Then hardware testing on both variants turned up two more (23065a4): on the classic, most lines started one column in, because the separator blank is emitted before the word length is known, so a line ending exactly on its last column pushes that blank onto the next line — the LCD has discarded a wrapped leading blank since V8, the OLED never did, as that block sat inside `#ifdef CONFIG_TFT`. It is unconditional now. And the look-ahead was extended to LoRa/WiFi receive and CW play, which have the same property.

Closing this one since the commits went in rebased rather than merged.

PR #204, opened 2026-08-12
Koch Lesson menu: show total lesson count alongside the current one

ckonecny:
## Summary
- Koch Trainer → Select Lesson now shows `N/M: x` (lesson N of the sequence's M total, then the newest character) instead of just `N char x`, so the total is visible without leaving the menu.
- `kochMaximum` already tracks the correct total for every sequence (native/LCWO/CW Academy/LICW carousel window/Custom Chars/adaptive), so no new state was added.
- Kept the format slash-compact (`N/M: x`) to stay inside the 14-char OLED value-line budget even for a worst-case prosign character like `<err>`.
- Updated the EN + DE user manuals (Version 8.x) to mention the new display format.

## Test plan
- [x] Builds cleanly for `heltec_wifi_lora_32_V2` (Classic, OLED)
- [x] Builds cleanly for `pocketwroom` (Pocket, LCD)
- [x] Flashed to a connected M32 Pocket and confirmed the menu shows e.g. `21/39: y`
- [x] Re-ran `extract_voice_strings.py` — empty diff, no accessibility voice clip owed (the Koch Lesson value line was already exempt as dynamic content)

### DIFF
--- a/Documentation/User Manual/Version 8.x/manual_de.md
+++ b/Documentation/User Manual/Version 8.x/manual_de.md
@@ -1000,8 +1000,9 @@ einzustellen.
 ### Koch: Select Lesson
 
 Wähle eine „Koch-Lektion" zwischen 1 und 51 (insgesamt 51 Zeichen
-werden nach der Koch-Methode gelernt). Lektion und zugehöriges Zeichen
-werden im Menü angezeigt.
+werden nach der Koch-Methode gelernt). Im Menü werden die Lektionsnummer,
+die Gesamtzahl der Lektionen der aktuell gewählten Koch-Sequenz (z. B.
+`21/39:`) und das zugehörige Zeichen angezeigt.
 
 Die Zeichenreihenfolge wurde von Koch nicht streng festgelegt, daher
 verwenden verschiedene Lernkurse leicht unterschiedliche Reihenfolgen.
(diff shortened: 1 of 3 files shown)


oe1wkl:
Merged to master as 2333241, together with a small follow-up (d41e1ce). Thanks — `kochMaximum` was indeed already correct for every sequence, so this needed no new state, and the value line even came out a column narrower than the old `N char x`.

Two things I changed on the way in:

**Manual → V9.** The edit had gone into `Documentation/User Manual/Version 8.x/`, which is the frozen released V8 manual; master is V9, so the live one is `Version 9.x/`. Moved there for EN + DE, HTML and PDFs rebuilt. I also corrected the surrounding claim that the lesson runs "between 1 and 51" — true only for the default M32 sequence (LICW BC1 has 18, BC2 runs 19–44, Custom Chars as many as the set has), which is exactly what your new total makes visible.

**Separator `/` → `-`,** so the menu now reads `21-39: y`. That value line is also emitted over the serial protocol as the last element of `getMenuPath() + "/" + value`, and the protocol defines menu path elements as slash-separated, so a `/` in the value split the leaf into two bogus elements for every lesson — for the config tool and the accessibility client that parse that stream.

One happy side effect: reviewing this surfaced that the Accessibility Edition never spoke this value at all (it was assembled at runtime, so no clip could match) — that gap was ours, not yours, and it is now fixed on master too. A blind operator hears "Koch lesson, 21 of 39, Yankee".

PR #205, opened 2026-08-17
Add adjustable scroll-area font size (Normal/Small)

ckonecny:
## Summary
- New "Font Size" preference (right after **Theme**) lets the scroll area — CW Generator/Echo Trainer/Keyer read-along, menus, and other screens sharing it — show a smaller, denser font for more characters per line. Also reachable via a triple click of the FN button.
- On the M32 Pocket, **Small** also fits a 5th visible line (4 normally). This uses a dedicated ASCII-only IntelOneMono variant (`IntelOneMono12ptAscii.h`) rather than the library's default 0x20–0xFF glyph range: LovyanGFX sizes and positions every glyph from the font's *whole* embedded range, and the accented characters this English-only UI never draws were inflating the line pitch just enough to block a 5th row. Reuses the exact same bitmaps — same look, narrower declared range.
  - Trade-off: the boot splash's "©" falls outside that range, so the splash is drawn at the normal size regardless of the preference (`forceNormal` param on `printOnScroll()`).
- Preference is modeled as a standard `pliste[]`/`prefPos` entry (`posScrollFont`), following the "Encoder Click" pattern — visible in every preferences menu Theme appears in.

## Scope (updated after review)
Per discussion with Willi outside GitHub: this is now **M32 Pocket, non-Accessibility only**. It doesn't touch the Accessibility Edition (purely visual, no benefit there) or the classic OLED variant (the small font asset wasn't designed to be a *default* reading size, not just narrow special cases — Willi flagged this specifically).

Implementation: a new build flag, `CONFIG_SCROLL_FONT_SIZE`, set for `pocketwroom` (and its `pocketwroom-lora`/`pocketwroom-170x240` derivatives), explicitly `build_unflags`-removed for `pocketwroom-accessibility`, and never defined on OLED (no `CONFIG_TFT`). Every part of the feature — the `prefPos` entry, its menu wiring, the ASCII-range font, the runtime 5-line geometry, the FN triple-click — lives behind that one flag. OLED's own code paths (font selection, wrap math, buffer sizing) are back to their pre-feature form.

Manuals: the "Font Size" preference and its FN-button mention are now Pocket-only text spans (same pattern already used for "Invader Orient.") — shown in Pocket manuals, excluded from the Accessibility variant, dropped entirely from the Classic edition sections. Re-ran the voice-string extractor to confirm "Font Size"/"Small" are no longer in the owed clip list for the Accessibility build.

## Also fixes (uncovered while building this)
- An OLED buffer-capacity constant had silently detached from the real per-line wrap width, letting a long generated word overflow the display edge.
- A menu screen cleared/redrew at the wrong font size when the row's own font changed mid-session, leaving stray pixels.
- An unspecified-evaluation-order dependency in one column-width calculation.

## Test plan
- [x] Builds clean: `heltec_wifi_lora_32_V2`, `pocketwroom`, `pocketwroom-accessibility`
- [x] Hardware-tested on M32 Pocket: FN triple-click and menu toggle, both directions, including boot with Small already saved
- [x] Menu entry appears in all preference views (top-level "All" and per-mode submenus)
- [x] Confirmed absent from `pocketwroom-accessibility` build and its voice-clip extraction
- [ ] OLED hardware testing (only build-verified — code paths are unchanged from pre-PR master, so no behavior to test there)
- Manuals (EN+DE) updated; PDFs not rebuilt (no local weasyprint) — TODO for release build


oe1wkl:
Thanks Christian — nice piece of work, and the scoping discussion we had outside GitHub is reflected cleanly in the code: one flag, `build_unflags` for the a11y edition, OLED paths back to their pre-feature form, variant-tagged manual spans, net-zero voice-string diff. The evaluation-order split in `refreshScrollLine()` is a real latent-bug catch that had nothing to do with this feature — good spot.

I do have one blocking finding plus a handful of smaller ones. Details below.

---

## 1. Blocking: the ASCII-range font never actually changes the line pitch

This is the mechanism the whole `IntelOneMono12ptAscii.h` exercise rests on, and I don't think it engages.

`setDefaultScrollFont()` and `printOnScroll()` select the small font through the **pointer** overload:

```cpp
display.setFont(&IntelOneMono_Regular12pt8b_Ascii);
```

In DisplayWrapper (current `oe1wkl/DisplayWrapper` master) that overload is:

```cpp
void DisplayWrapper::setFont(const GFXfont *fontData){
	lcd.setFont(fontData);          // note: no  currentFont = fontData;
}
```

whereas the **index** overload (`setFontIndex()`) does assign `currentFont`. And:

```cpp
uint8_t DisplayWrapper::getStringHeight(const String& strUser){
	return lcd.fontHeight(currentFont);
}
```

Since `LINE_HEIGHT` on TFT is `display.getStringHeight("j")`, every row's Y position is computed from **whichever font was last selected by index**, not from the ASCII font actually being drawn. Width is fine (`getStringWidth` / `textWidth` use the live LGFX font), so `printOnScroll()` currently measures X in the new font and Y in the old one.

I computed the four fonts' LovyanGFX default metrics from the glyph tables — they match your header comment exactly:

| font | `H(0x20–0xFF)` | `H(0x20–0x7E)` | xAdvance |
|---|---|---|---|
| IntelOneMono 12pt | 29 | **25** | 15 |
| IntelOneMono 15pt | 35 | 30 | 18 |

So in practice the pitch is 29 (the status line's `setFont(DialogInput_plain_12)` leaves `currentFont` at the full-range 12pt), never the 25 the ASCII range was created to produce. Five rows then span `29 + 5×29 = 174` px in a 141 px area — it only *looks* right because the last row's glyphs are 25 px tall and end at exactly 170.

A concrete way to see it: **boot with Font Size = Small and watch the splash.** `displayStartUp()` draws lines 0/1 with `forceNormal` → index 2 → `currentFont` = the 15pt font (35). `displayBatteryStatus()` then calls `printOnScroll(NoOfVisibleLines-1 = 4, …)` with the pointer-selected ASCII font, so `LINE_HEIGHT` is still 35 and `y = 29 + 4×35 = 169` on a 170 px panel — the `U: x.x V` line is off the bottom edge.

**Recommendation:** fix it in DisplayWrapper — it's one line in a fork Willi controls:

```cpp
void DisplayWrapper::setFont(const GFXfont *fontData){
	lcd.setFont(fontData);
	currentFont = fontData;
}
```

I checked: no existing firmware code calls the wrapper's pointer overload (every `setFont(&…)` in the tree is on a `canvas` sprite, not on `display`), so this can't regress anything today — it just closes a trap for the next caller. With it, the pitch really becomes 25, `29 + 5×25 = 154` leaves ~16 px of margin instead of 0, and the splash line lands at 129.

If you'd rather not touch the library in this PR, the alternative is for `MorseOutput` to stop asking the wrapper for the height and own the scroll-area pitch itself — but that's a bigger change and I'd prefer the one-liner.

### Mechanics — I'll take the library side

Touching DisplayWrapper has more process attached to it than the one line suggests, and most of that process is mine rather than yours. **So: I'll own the library change, the upstream PR and the hardware re-test. Please leave `setFont()` alone in this PR and treat the wrapper fix as arriving separately.**

What that leaves on your side is everything from §2 down, plus one decision: whether you'd rather hold this PR until the library fix lands (cleanest — the layout only becomes correct once both halves are in), or land the firmware side first knowing the pitch is still coming from the wrong font until I catch up. I'd lean towards holding, but it's your call and I don't want to block you for long.

For the record, here's what I'll be dealing with.

**Upstream.** What we build against is my fork, `oe1wkl/DisplayWrapper`, of `haklein/DisplayWrapper`. I diffed them, and the divergence is almost nothing:

* `getLGFX()` is unconditional in the fork; upstream still gates it behind `#if defined(ST7789_DRIVER) || defined(ILI9341_DRIVER)` (which our Pocket builds define anyway, so it isn't load-bearing for us)
* re-indentation of `drawString()` plus a commented-out older version of it, and a stray `src.code-workspace`
* `library.json` says 1.0.1 in the fork, 1.0.3 upstream — the fork simply predates Harald's version bumps

So there is no Morserino-specific divergence worth protecting, and the `currentFont` fix is a plain bug in the wrapper (the pointer overload silently desynchronises `getStringHeight()`), not something specific to us — I'll send it to Harald as a PR rather than keeping it fork-only. Long term the fork may not need to exist at all.

**Pinning.** `lib_deps` currently references the fork with no revision:

```ini
https://github.com/oe1wkl/DisplayWrapper   ; Re-added (was removed in #157) …
```

Every clean build pulls whatever the default branch happens to be at that moment, and no firmware revision records which library revision it was built against. That's tolerable while the library never changes; it stops being tolerable the moment we start changing it. PlatformIO accepts a revision on a VCS URL, so I'd like us to pin before the first change lands:

```ini
https://github.com/oe1wkl/DisplayWrapper#v1.0.4
```

and then bump the pin deliberately. That also gives us a rollback that doesn't involve force-pushing a library branch.

**The unlit-TFT-on-USB-only issue.** This is the one that worries me most. The reason the library dependency exists in its current form at all is #157 → #174: replacing it with the in-tree `M32PocketLGFX` left the panel dark on Pockets running from USB alone, ~19 fix-forward attempts failed, and the root cause was never pinned down. A source-level change inside the existing library keeps the library-archive build path intact, and `setFont()` is nowhere near `init()` or the power sequencing, so on paper this shouldn't go anywhere near it. But "on paper" is exactly what we thought last time.

So before the library change goes anywhere near master I'll re-test a cold boot on USB power **with the battery physically disconnected**. A battery in the loop masks the failure, which is how it got shipped the first time. This is the main reason I want the library half on my bench rather than in this PR.

**And the layout genuinely moves.** Once the fix takes effect the pitch really becomes 25 px, so every scroll row shifts up and the Pocket screens want another pass on hardware — this isn't a fix we can land on a build check alone.

---

## 2. Two more places change the preference without re-deriving the geometry

`applyScrollFontGeometry()` is wired into boot, `toggleScrollFont()` and the menu adjust — but `pliste[posScrollFont].value` can also change from:

* **`applySnapshot()`** (`MorsePreferences.cpp`) — `storedInSnapshot()` returns `true` for it via the default branch, and the function ends at `handleKochSequence(); updateTimings();`
* **`setParameter()`** (`m32_v6.ino`) — the protocol `PUT parameter/font size/…` path, which has special cases for `posKochSeq` / `posCarouselStart` / `posSidetoneShape` but not this one

Either way the font changes and `NoOfVisibleLines` / `maxPos` don't. Recalling a *Normal* snapshot while running Small leaves 5 visible lines at a 35 px pitch: line 4 lands at y = 169 and the newest scroll line is invisible.

Theme survives both paths only because `writePreferences()` re-applies it, and both of these call `writePreferences()` — so I'd put the hook right next to the theme one:

```cpp
#ifdef CONFIG_TFT
      if (pliste[posTheme].value != pref.getUChar("theme")) { … }
#endif
#ifdef CONFIG_SCROLL_FONT_SIZE
      MorseOutput::applyScrollFontGeometry();
#endif
```

One thing to change first: `applyScrollFontGeometry()` currently does `relPos = maxPos`, i.e. it jumps the view to the bottom. That's fine for the FN toggle but wrong as a general sync hook — better as

```cpp
relPos = constrain(relPos, 0, maxPos);
```

with the explicit `relPos = maxPos` left where the toggle wants it.

**A tempting shortcut that I don't think we should take:** you could sidestep the snapshot half entirely by adding `case posScrollFont: return false;` to `storedInSnapshot()`, on the grounds that Font Size is device/UI appearance like `posClicks` and `posQuickStart`, which are excluded.

I'd have suggested exactly that, but Theme being in snapshots is a **deliberate** decision, not an oversight: carrying the colour scheme into the preset means the display tells you at a glance which snapshot you are on. That's a real feature, and Font Size can serve the same purpose — a denser screen is as recognisable as a different palette. So the precedent points the other way, and I'd rather we keep Font Size in snapshots and fix the geometry sync properly via the `writePreferences()` hook above.

Worth a short comment next to `storedInSnapshot()` recording that rationale, incidentally — the function's own doc comment currently reads as "device settings are excluded", which is what led me down the wrong path in the first place.

---

## 3. Toggling Small → Normal clips the existing scrollback

`toggleScrollFont()` repaints the buffer with `refreshScrollArea()`, but those lines were wrapped at the *old* column count. Small wraps at 320/15 = 21 columns; redrawn at 18 px/char, column 18 lands at x = 324 on a 320 px panel, and `DisplayWrapper::drawString()` doesn't re-wrap — so the last 3–4 characters of every buffered line are simply clipped away, including while scrolling back through the 18-line history.

Re-wrapping the buffer is a big change and I don't think it's worth it. Two reasonable ways out — your pick:

* clear the scroll buffer on toggle (`MorseOutput::clearScroll()`), so the user starts fresh at the new geometry rather than reading truncated history; or
* leave it and say so in the manual ("history already on screen is redrawn at the new size and may be truncated").

---

## 4. Smaller items

* **`writePrefPosNow()` doesn't check its NVS results.** CLAUDE.md §4(b): *"check `put*()`/`begin()` results — they fail silently when NVS is full (error logs are compiled out at `CORE_DEBUG_LEVEL=0`)"*. On a device near the 504-entry ceiling the toggle would appear to work and silently revert at the next boot. `doWriteSnapshot()` in the same file shows the pattern.
* **`MorseOutput::clearLine(2)` in `displayValueLine()` is unconditional — and it's the one thing in this PR that reaches the classic build.** It fixes a problem that only exists for the single preference whose adjustment changes the row's own font, but it has no `#ifdef` around it, so it now runs for all ~40 preferences on every encoder click, on **both** variants.

  On the Pocket that's an extra full-width panel write per click, on a path we already had to special-case for flicker (the `statusClearPending` cache) and with Phase G still open.

  On the classic OLED I'd expect it to be invisible in practice — it's a framebuffer write flushed by the `display.display()` inside `printOnScroll()`, so no flash, and the only pixels it newly touches are column 0 of that row, which the preferences menu leaves blank. But it does mean the test plan's justification for skipping OLED hardware testing — *"code paths are unchanged from pre-PR master, so no behavior to test there"* — isn't quite accurate as written. Everything else that reaches the classic build really is neutral: `setDefaultScrollFont()` collapses to the same `setFont(DialogInput_plain_15)`, `forceNormal` is ignored in the `#else` branch, and the `refreshScrollLine` split is a no-op there because `C_WIDTH` is the constant `9`, not a font query. The only other residues are `maxPos` losing its `const` on both variants and `writePrefPosNow()` being compiled into the classic image but never called there.

  Guarding the clear with `#ifdef CONFIG_SCROLL_FONT_SIZE` and `pos == posScrollFont` — or folding it into the `adjustKeyerPreference` branch that already repaints row 1 — fixes the Pocket cost, keeps the classic build genuinely untouched, and makes that test-plan line true as written.
* **`toggleScrollFont()` leaves the scroll bar stale.** It resets `relPos` and changes `maxPos`, but the wipe is `fillRect(0, SCROLL_TOP, getWidth()-1, …)` which spans x 0…318 and doesn't cover the bar at x = 319 — so in scroll-back mode the old thumb stays painted at its old offset. A `displayScrollBar()` refresh (or dropping out of `scrollMode`) closes it.
* **CLAUDE.md §3 rule 6 is now stale** — *"Visible scroll lines are governed by the `NoOfVisibleLines` **constant**"*. It's a runtime `uint8_t` on the Pocket now; worth updating in this PR so the next change doesn't get made against the old rule.
* **Manual: missing space before the `.pocket` span.** `steps.[ A triple click …]{.pocket}` — pandoc drops the leading space, and the committed HTML already shows `several steps.<span class="pocket">A triple click` (and `in mehreren Stufen.<span class="pocket">Ein Dreifachklick`). The convention elsewhere in the file keeps the space outside the bracket, e.g. `the [LoRa or]{.classic} WiFi Transceiver modes`. Both EN and DE.
* **Header comment nit.** `IntelOneMono12ptAscii.h` says `getGlyph()` "indexes into the glyph array as (codepoint - first) regardless of `last`" — it actually does range-check (`if (uniCode > last || uniCode < f) return nullptr;` in `lgfx_fonts.cpp`), which is precisely *why* the `©` comes out as a blank rather than a wrong glyph. Your conclusion is right, the stated reason isn't, and that comment is load-bearing.
* **Duplicated font choice.** `setDefaultScrollFont()` and the new `#ifdef` block in `printOnScroll()` both map the preference to a font, differing only in the bold dimension. A single `scrollFont(bool bold, bool forceNormal)` that `setDefaultScrollFont()` calls with `(false, false)` keeps them from drifting — and if we take the DisplayWrapper fix above, there's exactly one place to touch.
* **Font header duplication.** The quoted includes resolve to the *in-tree* `IntelOneMono_{Regular,Bold}12pt8b.h` (same-directory wins), not DisplayWrapper's. They're byte-identical today and nothing in-tree included them before, but "same bitmaps" now depends on two independently maintained copies staying in sync across library updates. A comment naming the in-tree copies as the source of truth would be enough.

---

Nothing here is a design objection — the flag layout, the preference modelling and the manual/a11y handling are all right. Item 1 is the one that has to be resolved before merge, since the current 5-line layout has no pixel margin and depends on which font was last selected by index; I'll take that half.

---

## Not for this PR, but you prompted the thought

Excluding Font Size from the Accessibility Edition as "purely visual, no benefit there" is the right call, and it makes me look again at **Theme**, which *is* still in that build on the same purely-visual reasoning. For a completely blind operator it earns nothing either. Where it could earn a lot is one step further out: proper **high-contrast themes** aimed at partially-sighted users, who are a real part of that audience and are currently served by neither the colour themes (tuned for looks) nor the speech (aimed at no vision at all). Large, high-contrast text plus speech is a different and probably more common profile than speech alone.

That's a product question for the a11y roadmap rather than anything you need to act on here — noting it so the thread doesn't lose it.

ckonecny:
Thanks for the thorough pass, Willi.

**§1 (blocking):** Agreed — let's hold this PR until your DisplayWrapper fix and the USB-only cold-boot re-test land. Landing the firmware side first with a known-wrong pitch and zero margin isn't worth it just to merge sooner. I'll leave `setFont()` alone.

**In the meantime, one more scope change:** we've dropped the FN triple-click gesture entirely (commit 9a23d06, pushed just before your review landed — sorry for the crossed wires). It turned out to be genuinely unstable: toggling it live while a mode kept writing to the scroll area (e.g. mid-word in the CW Generator) desynced the write cursor's column tracking between the old and new font's metrics and corrupted the bottom line. The preferences menu and the serial config UI already reach the same setting without that race, since nothing writes to the scroll area while its own menu is open — so removing the redundant runtime gesture eliminates that bug class rather than working around it, and keeps the overall solution simpler. `toggleScrollFont()` is gone; `applyScrollFontGeometry()` stays (boot + menu encoder-adjust).

That also resolves a couple of your findings directly:
- §3 (scrollback clipping) as filed against `toggleScrollFont()` is moot — that function no longer exists. The underlying mechanism (`refreshScrollArea()` redrawing old-wrapped lines at the new font's pixel width) may still be reachable via the menu path when returning to a mode after changing Font Size there; I'll check that before calling it closed.
- §4 scroll-bar staleness in `toggleScrollFont()` — moot, whole path deleted.
- §4 missing space before the `.pocket` span in the manual's device-overview row — moot, that sentence is gone along with the gesture.

Working through the rest of §2–§4 now (geometry re-sync via `writePreferences()`, the `relPos` constrain fix, `writePrefPosNow()`'s unchecked NVS result, guarding `clearLine(2)` so it stops reaching the classic build, the doc/comment nits) plus one you couldn't have caught: `writePrefPosNow()` was `toggleScrollFont()`'s only caller, so it's dead code now — cleaning that up too, most likely by removing it in favor of the `writePreferences()` hook rather than keeping a now-purposeless function around.

Will report back here once that's done and ready for another look, ahead of your library fix landing.

ckonecny:
Pushed (commit 20aca5d) — everything from §2–§4 except what's yours to take (§1) or genuinely not worth doing (re-wrapping the buffer in place for §3, see below).

**§2, geometry re-sync:** Added a `posScrollFont` case to `writePreferences()`'s generic `pliste[]` loop, so `applyScrollFontGeometry()` now runs for snapshot recall and the serial protocol too, not just the menu's live-preview handler — one hook covers every path that can change the value. Also switched it to `relPos = constrain(relPos, 0, maxPos)` as you suggested, with the explicit jump-to-bottom now only at boot (where the buffer is empty, so it's unambiguous). Also added the rationale comment you asked for next to `storedInSnapshot()` — Font Size follows Theme's precedent (kept in snapshots on purpose, not an oversight).

**§3, scrollback clipping — this one I actually re-checked rather than assuming your original finding was moot.** You're right that `refreshScrollLine()` still draws `textBuffer[]` content at whichever font is *currently* active, not the one it was wrapped under when written — that mechanism is untouched by removing the FN gesture. I had a subagent trace all three remaining ways to change the value: turns out it's not reachable from the menu-exit path in Keyer/CW Generator/Echo Trainer (menu-exit there never redraws the buffer, it just wipes the screen), and not reachable from a bare serial `setParameter()` either — but it *is* still live in QSO Bot mode, which calls `refreshScrollArea()` right after its own preferences-menu return with `posScrollFont` reachable from `qsoBotOptions`. And more generally, any mode's later manual scroll-back through history buffered before a Font Size change hits the same thing. So this wasn't a false alarm from the old toggle — it's a real, still-open gap, just with different trigger conditions than what you originally filed it against. Took your first option: `MorseOutput::clearScroll()` (already existed, used at mode transitions) now fires whenever `writePreferences()` sees the value actually change. Settings change, scroll area starts clean — didn't attempt re-wrapping in place, agreed that's not worth it here.

**§4:**
- `writePrefPosNow()` removed — dead code, was the deleted toggle's only caller, and NVS persistence for this preference is now handled entirely by the generic mechanism.
- The unconditional `clearLine(2)` in `displayValueLine()` — turned out to be flat-out redundant, not just unguarded: the adjacent `clearScrollLines()` call in the menu's `posScrollFont` branch already clears that row (it loops every visible line). Deleted it outright rather than adding a guard, so `displayValueLine()` is now byte-identical to upstream again.
- Consolidated the duplicated font-selection logic into a shared `scrollFont(bool bold, bool forceNormal)` that both `setDefaultScrollFont()` and `printOnScroll()` call — also let me drop `printOnScroll()`'s now-pointless `#ifdef`/`#else` split entirely, since both branches had converged to the same three lines.
- Fixed the `getGlyph()` comment in `IntelOneMono12ptAscii.h` (range-check, not unchecked indexing — your correction) and added the in-tree-copies-are-source-of-truth note.
- Updated CLAUDE.md §3 rule 6 off "constant."

Rebuilt and confirmed clean on all three environments, and confirmed the OLED build is genuinely untouched again (`git diff upstream/master` shows zero residual diff on `displayValueLine()`), and the voice-string extractor still reports a net-zero diff.

That's everything on my side ready to go whenever your library fix and hardware re-test land.

ckonecny:
Hardware-tested end to end on commit 6d48ccd (M32 Pocket):
- Font Size Normal↔Small via the preferences menu, both directions
- Scroll buffer clears cleanly on an actual change instead of showing clipped old text
- Snapshot recall while running: save a snapshot with one Font Size, switch the live setting to the other, recall the snapshot mid-mode — display and live text both track correctly

Also found and fixed one more bug during this testing pass, not part of your original review: my §2 `relPos = constrain(relPos, 0, maxPos)` fix broke the far more common case — a view actively glued to the live bottom. Going Small→Normal increases `maxPos`; the old `relPos` (== old, smaller `maxPos`) was still within the new range, so `constrain()` left it numerically unchanged, which only happens to equal the new `maxPos` in the *other* direction. Once `relPos != maxPos`, the CW Generator's live-print fast path stops firing — text keeps generating, the screen just stops updating, silently. Fixed in 6d48ccd by capturing "was glued to bottom" before recomputing geometry and landing `relPos` on the new `maxPos` explicitly in that case, rather than trusting `constrain()` to land there by coincidence.

Everything on my side is tested and ready. Still waiting on your DisplayWrapper fix + USB-only cold-boot re-test for §1 before this can merge.

oe1wkl:
Item 1 is cleared — the library half is done and on master, so this is unblocked.

**What landed** (`master`, two commits):

* `bc2379d` — `lib_deps` now pins DisplayWrapper to an exact revision instead of a bare URL. Unpinned, every clean build pulled whatever the fork's default branch happened to be, which is not something I want for this particular dependency.
* `103e9f4` — bumps that pin to the fix: `DisplayWrapper::setFont(const GFXfont *)` now updates `currentFont`, so `getStringHeight()` (and therefore `LINE_HEIGHT`) reports the font that is actually selected. There is also a NULL guard for a `getStringHeight()` before the first `setFont()`, which would previously have faulted.

Hardware-checked before landing, since this is the dependency behind the #157 dark-panel business: M32 Pocket, battery switch off, USB power only, three cold boots — panel lit and splash rendered every time. Both canonical variants build. Harald has since merged the same fix upstream (haklein/DisplayWrapper#3).

Nothing in master calls that overload, so the bump changes no firmware behaviour on its own — it only matters once your branch uses it.

**Please rebase onto master before testing again.** On the current base the pitch still comes from whichever font was last selected by index, so the five-line layout appears to work for the wrong reason and with no pixel margin. Rebased, the small font really does report 25 px, which is what the ASCII-range header was for: `29 + 5×25 = 154` on a 170 px panel, ~16 px to spare. Expect every scroll row to shift up a little compared to what you tested — that is the fix taking effect, not a regression.

The rest of the review still stands and is all firmware-side:

* the geometry sync for snapshot recall and the protocol `PUT parameter` path (§2) — with `applyScrollFontGeometry()`'s `relPos = maxPos` changed to a `constrain` first, if you hook it into `writePreferences()`
* the Small → Normal scrollback clipping (§3)
* the NVS result check, the unguarded `clearLine(2)`, and the manual spacing (§4)

On §2, one correction to my own suggestion: I floated excluding Font Size from snapshots as a shortcut. Willi has since confirmed Theme is in snapshots deliberately — the colour scheme is meant to tell you at a glance which snapshot you are on — so that precedent points the other way. Keep Font Size in snapshots and fix the sync properly.

Thanks for your patience with the length of the first review; the interesting part turned out not to be in your code at all.

ckonecny:
Rebased onto master (commit 6dfc096) and retested on M32 Pocket hardware with your fix in place.

**§1 confirmed fixed.** After the rebase, the small font really does report the honest ~25px pitch instead of the stale ~29px — verified via the boot splash: battery status line ("U: x.x V") is fully visible in both Normal and Small, no clipping.

That surfaced something worth flagging: with the real (smaller) pitch now in effect, 5 tightly-packed rows left ~16px of genuinely new slack that previously didn't exist in the (wrong) math. All of it was landing as one lump below the last row — readable, but visibly top-heavy. Fixed by deriving a small inter-row gap plus a matching top margin from that leftover, split into equal shares so the block is centered top-to-bottom (not just "some padding added somewhere") — `applyScrollFontGeometry()`. Fully inert everywhere `NoOfVisibleLines` is a compile-time constant (OLED, TFT without the preference), so no diff there.

That, in turn, uncovered one more bug: `clearScrollLines()` used to clear N individual per-row bands, which was fine when rows were packed with zero gaps, but once gaps exist between rows they went uncleared — a fragment of the previous font's menu-preview text could sit exactly in a gap and never get wiped switching Normal→Small in the preferences menu. Switched to a single rect over the whole scroll area instead.

Hardware-tested end to end: boot in both sizes, CW Generator live output in both sizes and both switch directions, Small↔Normal via the menu with no leftover fragments, snapshot recall mid-mode. All clean.

Ready for another look whenever you have a chance.

PR #206, opened 2026-08-18
Add VK/ZL region option for callsign generator

VK1HF:
Adds a 'VK/ZL' option (value 7) to the Calls Region setting in the callsign generator. This allows Australian and New Zealand operators to practice with locally relevant callsigns rather than the broader Oceania region.

The generator produces realistic VK/ZL callsigns weighted by call area and population:
- 85% VK / 15% ZL split
- VK areas weighted by activity (VK2/VK3 most common, VK0/VK9 rare)
- ZL areas: ZL1/ZL2 35% each, ZL3 20%, ZL4 10%
- 70% three-letter suffixes, 30% two-letter suffixes

This was motivated by observing that the existing Oceania filter produced unrealistic callsigns for VK/ZL operators. Documentation updated in manual_en.md.

Changes:
- MorsePreferences.cpp: added 'VK/ZL' as 8th option in Calls Region menu
- morsedefs.h: added posCallContinent value 7 for VK/ZL
- m32_v6.ino: custom weighted VK/ZL callsign generator bypassing prefix table
- manual_en.md: documented new option


oe1wkl:
Thanks Ian — the motivation is real and I'd like to get proper VK/ZL practice into V9. A few things need sorting before it can go in.

### 1. The branch is stale, and its first commit is already on master

This PR branches from `0536404`, which predates `c50e0a0` ("Add VK/ZL region option for callsign generator (#177)", merged 2026-06-01). Your first commit `d917985` *is* that work, so the diff shown here re-adds the preference entry, the `maximum` bump and the two `strncmp` filters that master already carries. Rebasing on master makes git skip `d917985`, leaving just "Improve VK/ZL callsign realism…" — which is the part I actually need to review. Could you rebase?

### 2. Missing string terminator — this corrupts call signs

The new branch ends with

```c
call[pos] = ' ';
```

where the table path uses `call[pos] = '\0';`. `call` is a `static char[16]` reused across invocations, so `String(call)` runs past the space into whatever the previous call left behind. Reproduced with a standalone harness using the same buffer:

```
after table path       : "dl1xyz/m"
VK/ZL, 3-letter suffix : "vk2abc m"      <- stale 'm' from the previous call
VK/ZL, 2-letter suffix : "vk2ab  m"
```

Even a pure VK/ZL run leaks a second trailing space whenever a 3-letter call is followed by a 2-letter one. Trailing spaces matter here — the decoder already emits one, and everything downstream trims on that assumption.

### 3. "Length Calls" is silently ignored

The early `return` happens before `maxLength` is used at all, so with Length Calls set to max. 3/4/5 you still get 5–6 character calls. The table path honours that preference; the new branch bypasses it, along with "Call Prefixes: Common only".

### 4. The bigger question: extend the table rather than fork the generator

Your diagnosis is correct — `callsign_prefixes.h` currently has `{"VK", 0x20, 29, 156}` and `{"ZL", 0x20, 32, 120}`, plus `VK0`/`VK9`/`ZL7`/`ZL8`/`ZL9`. That works out to roughly a 55/45 VK/ZL split with a uniform-random area digit, which is indeed not what the bands sound like.

But the table already supports per-area rows — `VK0`, `VK9`, `ZL7`, `ZL8` and `ZL9` are in there today. Adding weighted `VK1`…`VK8` and `ZL1`…`ZL4` entries and downweighting the bare `VK`/`ZL` rows would give you the 85/15 split and the area distribution you want, and would do it *inside* the existing machinery:

- Length Calls and "Common only" keep working automatically.
- CQ zones come from the table per area instead of a hardcoded `isVK ? 29 : 32` (which also loses the table's `VK0` → 30 and `VK9` → 28, and those zones feed the QSO Bot's CQ WW exchange).
- Oceania and All improve too, not just the new VK/ZL option.
- No second call-sign code path to keep in step.

Mechanically it fits: a 3-character prefix ending in a digit skips the extra-digit step, so a `VK2` row yields `vk2` + suffix as you'd expect. The header says auto-generated, but no generator script is committed, so the table is hand-maintained in practice and new rows are safe to add.

That's the route I'd prefer (the project convention is to extend an existing helper rather than fork it). If you'd rather keep the dedicated branch, it needs the `'\0'` fix, `maxLength` handling, and removal of the two `vkzlOnly && strncmp(...)` lines in the prefix passes — they became unreachable the moment the early return went in.

### Accessibility and documentation — I'll take these

Two housekeeping duties that a change like this normally carries, both of which I'll handle so you don't need the toolchain:

- **Voice clips:** nothing owed here, as it happens. I ran the extractor against your rebased branch and the diff is empty — "VK/ZL" already got its clip when #177 landed.
- **Manual:** your change targets `Documentation/User Manual/Version 8.x/manual_en.md`, which is frozen. The live manual is `Version 9.x/manual_en.md` and `manual_de.md`, and it turns out neither of them mentions VK/ZL at all — #177's documentation went into V8 only and was never carried forward. I'll write the English and German entries against the V9 sources.

So: a rebase plus the code fixes from your side, and the manual from mine.

oe1wkl:
Anything to be expected from your side?

VK1HF:
I've been so snowed under with work Willi - will try to get on to these
both!

Regards,

Ian

oe1wkl:
Merged — thank you, Ian. Your commits are on `master` (merge `ab343cf`), and the weighting is exactly what that option needed: picking the call area uniformly made VK8 as likely as VK2, which is not what anyone hears on the air.

Three things changed on the way in, all small:

- **The V8 manual is frozen**, so I reverted that hunk and wrote the row for the V9 manual instead, in English and German. It turned out `Calls Region` had never been documented at all — the setting has shipped since V8.1 — so the new row covers the whole preference, VK/ZL included.
- **`maxLength == 1`** ("Length Calls: max. 3") reached the VK/ZL branch and got a 5- or 6-character call back. Every other region is deliberately ignored for 3-character calls, and the manual says so, so VK/ZL now follows the same rule — there is no 3-character VK or ZL call to give. With that, the two `vkzlOnly` filters in the prefix-table passes became unreachable and are gone.
- **The terminator** at the end of the branch was a raw NUL *byte* in the source rather than the `'\0'` escape. It compiled to the right value, so nothing misbehaved on the device — but it made `m32_v6.ino` binary as far as `grep`, `diff` and several editors are concerned, silently, for the whole file. Worth knowing about: it is easy to introduce from an editor or a paste and hard to notice afterwards.

The Accessibility Edition already had a "VK/ZL" clip but nothing telling it how to pronounce it; it now speaks "Australia and New Zealand".

One question back to you, as the person best placed to answer it: `lastGeneratedCallCqZone` follows the prefix table, which puts plain `VK` in CQ zone 29 — but most of VK is zone 30, with 29 being VK6 and VK8. That is a question about the table rather than about your change, so I have left it alone. Worth correcting for the QSO Bot's CQ WW exchange?

VK1HF:
Re "Worth correcting for the QSO Bot's CQ WW exchange?" I think zone 29 is
fine. if it bothers users we could just change it at that point in time,
for the moment I say leave it how it is.

PR #207, opened 2026-08-18
Add Echo Think Time preference for head copy practice

VK1HF:
Adds a new 'Echo Think Time' preference to the Echo Trainer, allowing operators to set a fixed number of seconds (0-20s) added after the prompt before the trainer expects a response.

This is particularly useful for head copy practice with call signs, where the operator needs thinking time to mentally reconstruct what they heard without having to reduce the prompt speed or change the dot/dash ratio — which affects the whole device globally, not just the Echo Trainer.

The existing workaround suggested by users (adjusting dot/dash ratio or inter-word space) changes global device behaviour. This change is completely isolated to the Echo Trainer state machine.

Changes:
- morsedefs.h: added posEchoThinkTime to preference enum
- MorsePreferences.cpp: added preference entry (0-20s) and included in all echo trainer option lists; added echoThinkTime to prefName array
- m32_v6.ino: applied think time to genTimer calculation in echo trainer state machine
- manual_en.md: documented new preference in prose section and reference table

### DIFF
--- a/Documentation/User Manual/Version 8.x/manual_en.md
+++ b/Documentation/User Manual/Version 8.x/manual_en.md
@@ -943,6 +943,12 @@ adaptive speed, the speed for keying your input stays at that maximum
 (so you can train your copying at faster speeds than you are comfortable
 to use when keying).
 
+**Echo Think Time:** Set this to the number of seconds you would like
+to wait after the prompt has been sent before the Echo Trainer expects
+your response. This is useful for head copy practice, especially with
+call signs, where you may need a moment to build a mental picture of
+what you heard before keying your answer. Default is 0 (no extra wait).
+
 **Intercharacter Space** and **Interword Space**: The first preference
 defines the pause between characters in the prompt the M32 is generating
 (the same as in the Generator modes, see there); both preferences also
@@ -2705,6 +2711,7 @@ time" after the prompt before you must begin keying your response!
 | Tone Shift | The pitch of the tone, when you are using the Echo Trainer mode or transmitting in a transceiver mode, can either be the same as the one you get from the receiver (or from the prompt in Echo Trainer mode), or can be a half tone lower or a half tone higher. | **No Tone Shift** / Up 1/2 Tone / Down 1/2 Tone |
 | Adaptv. Speed | If this is set to ON, the speed will be increased by 1 WpM whenever you gave a correct response in Echo Trainer mode, and will be decreased by 1 whenever you made a mistake. | ON / **OFF** |
 | Echo Speed Max | Set this to the maximum speed you would like to use for keying in Echo Trainer mode. If you set the speed faster than this value with the Encoder, or if it is getting faster as a result of adaptive speed, the speed for keying your input stays at that maximum. | **No limit** / 5 – 50 wpm (in steps of 5 wpm) |
+| Echo Think Time | Sets the number of seconds added after the prompt before the Echo Trainer expects your response. Useful for head copy practice with call signs, where you need thinking time without changing the keying speed. | **0s** / 1s – 10s / 15s / 20s |
 
 
 
--- a/Software/src/Version 6 and newer/MorsePreferences.cpp
+++ b/Software/src/Version 6 and newer/MorsePreferences.cpp
@@ -71,6 +71,7 @@ const char * prefName[] = {
                       "randomLength", "callLength", "callContinent", "callCommon","abbrevLength", "wordLength",
                       "GeneratorDispl", "wordDoubler", "echoDisplay", "echoRepeats", "echoConf",
                       "KeyExternalTx", "LoraCwTransmit", "goertzelBW", "speedAdapt", "echoSpeedMax",
+                      "echoThinkTime",
                       "KochSeq", "carouselStart", "latency", "randomFile", "extAudioOnDecod", "timeOut",
                       "quickStart", "outputCase", "autoStop", "maxSequence", "LoraChannel",
 #ifdef CONFIG_BLUETOOTH_KEYBOARD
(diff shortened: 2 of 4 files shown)


oe1wkl:
Thanks Ian — this is a well-targeted change and I want it in V9. The mechanism is right: `genTimer` is the deadline for *starting* to key (`doPaddleIambic` flips to `EVAL_ANSWER` when it expires), so extending it really does buy thinking time without touching anything global, and both `SEND_WORD` and `REPEAT_WORD` pick it up. It rebases cleanly on master and builds for both variants.

Two things need fixing, and then two duties I'll take off your hands.

### 1. The 15 s and 20 s settings don't do what they say

`pliste[].value` is the *index* into `mapping[]`, not the value itself — the serial protocol reports it as `mapping[value]`, and the menu displays the same. So with

```c
0, 0, 12, 1,
{"0s","1s","2s","3s","4s","5s","6s","7s","8s","9s","10s","15s","20s"}
```

and `pliste[posEchoThinkTime].value * 1000UL`, indices 0–10 happen to coincide with seconds, but index 11 shows "15s" and waits 11 s, and index 12 shows "20s" and waits 12 s.

The fix I'd suggest is to drop the mapping entirely:

```c
{
  0, 0, 20, 1,
  "…",
  "Seconds to wait before echo trainer expects your response",
  false,
  {}
},
```

Index then *is* seconds, so the class of bug disappears rather than being patched, and the menu shows 0–20 as a plain number. There's precedent — `Max # of Words` (`posMaxSequence`) is an unmapped numeric. It also matters for accessibility (below): the mapped version needs 14 new voice clips, the unmapped one needs 1.

Worth knowing either way: `MAX_MAP_ELEMENTS` is 15 and your mapping already uses 13, so filling the ladder out to a full "0s"…"20s" isn't an option.

### 2. The label is one character too wide for the classic M32

`displayKeyerPreferencesMenu()` prints the label into a 14-character line (`NoOfCharsPerLine` is 14 on the OLED). "Echo Think Time" is 15, so the last character falls off the right edge on the classic M32.

Every preference label that is *not* behind a config flag is currently ≤14 characters; the two longer ones — `Invader Orient.` and `Headphone Output` — are behind `CONFIG_CW_GAME` and `CONFIG_SOUND_I2S`, i.e. Pocket-only. This would be the first over-wide label on the OLED. Something like `Echo Think T.` or `Think Time` would fit; the full wording can live in the `spokenName` field, which exists precisely so a cramped display label doesn't have to be the spoken one.

### 3. Accessibility — I'll do this one

Every new preference and option value has to be given a voice clip for the Accessibility Edition, or it is simply silent for blind operators. I ran the extractor against your rebased branch:

- **As written:** 14 new clips — `Echo Think Time` plus `0s`…`10s`, `15s`, `20s`. The bare "0s"/"10s" spellings would also need pronunciation overrides, the way `1 ms` → `1 millisecond` and `10 wpm` → `10 words per minute` do, otherwise the voice reads "zero ess".
- **With the unmapped fix above:** exactly 1 new clip, the label itself — the integers 0–60 already exist as atoms.

This needs piper plus the voice models and ffmpeg, which you almost certainly don't have set up, so leave it with me. I'll add a `spokenName` carrying the unit so it announces as something like "echo think time in seconds" regardless of how short the display label ends up.

### 4. Manual — also mine

Your change targets `Documentation/User Manual/Version 8.x/manual_en.md`, which is frozen. The live manual is `Version 9.x/manual_en.md` and `manual_de.md`, and both need the prose paragraph and the reference-table row. I'll write them.

So: the arithmetic and the label from your side, the clips and the manual from mine.

VK1HF:
HI Willi - I must confess - I made these changes for myself primarily, for my own use cases - but I thought others would benefit - hence pushing them upstream to you. These changes I made, work perfectly on my M32 pocket. I note your comments. Now - more importantly - I used Claude to help me with these. Essentially I steered claude to achieve the outcome I wanted. Got to say, I was super impressed. I can code, but I am not a developer as such and do not appreciate the full code set for morserino. So my question is - should I keep working with my own skills (using Claude also) to respond to what you have said - or would you prefer I don't do that? I'm happy either way mate. I just share these back, if you want to say thanks but no thanks - I am 100% fine with that. Love your work on the project! - Ian

oe1wkl:
Ian, please continue, with the help of Claude. I am using Claude myself, and it speeds up the development process considerably.It helps to use a powerful model like Opus, and to point it to CLAUDE.md in the repository, so it learns about the rules and conventions quickly.Willi Von meinem Mobiltelefon gesendetAm 20.08.2026 um 02:26 schrieb VK1HF ***@***.***>:﻿VK1HF left a comment (oe1wkl/Morserino-32#207)
HI Willi - I must confess - I made these changes for myself primarily, for my own use cases - but I thought others would benefit - hence pushing them upstream to you. These changes I made, work perfectly on my M32 pocket. I note your comments. Now - more importantly - I used Claude to help me with these. Essentially I steered claude to achieve the outcome I wanted. Got to say, I was super impressed. I can code, but I am not a developer as such and do not appreciate the full code set for morserino. So my question is - should I keep working with my own skills (using Claude also) to respond to what you have said - or would you prefer I don't do that? I'm happy either way mate. I just share these back, if you want to say thanks but no thanks - I am 100% fine with that. Love your work on the project! - Ian

oe1wkl:
Ian, are you updating the PR as recommended, or do you prefer that I take that on?

VK1HF:
you are welcome to take them in if you like - and have time iof course. The
key thing is I wanted to share back the ideas which I have found helpful.

Regards,

Ian

oe1wkl:
Closing this in favour of #213, which is merged — but the feature and the design are yours, Ian, and the change history and the manual both say so.

Christian rebuilt it against current `master` (this branch predates 349 commits, including work touching the same preference arrays) with the two review points folded in: the value is unmapped 0–20, so the number shown is the number of seconds waited, and the label fits the OLED's 14-character budget as "Echo Think T.", with the full wording carried in `spokenName` for the Accessibility Edition.

It is in V9 now, documented in English and German, and the Accessibility Edition speaks it. Thank you for building it, and for the patience — head-copy practice with call signs is exactly the case it earns its keep in.

PR #209, opened 2026-08-24
Fix Koch lesson filter being ignored for Custom Character Sets

ckonecny:
## Summary
- With a Custom Character Set active, `Koch::setup()` built the CW Abbrevs / English Words / Mixed word-index tables against the **entire** custom set instead of the current lesson (`kochFilter`), so those modes could serve words containing characters far beyond the learner's current lesson.
- `writePreferences()` also skipped `koch.setup()` on lesson change whenever a custom set was active, so Adaptive Random's probability table went stale on every lesson advance too.
- Both fixed to always use `kochFilter`, matching how the rest of the Custom Chars code (`setCustomChars()`, `getNewChar()`) already treats it as the lesson boundary.

Affects Koch Trainer's CW Generator and Echo Trainer (Words/Abbrevs/Mixed/Adaptive Random) when a Custom Character Set is configured. Random mode and the TFT games (which read the live character set directly rather than through the filtered index tables) were already correctly bounded and are unaffected. Built-in Koch sequences (non-custom) were also unaffected, since they already passed `kochFilter` correctly.

## Test plan
- [x] `pio run -e pocketwroom` — builds clean
- [x] `pio run -e heltec_wifi_lora_32_V2` — builds clean
- [x] Flashed to an M32 Pocket and confirmed English Words / CW Abbrevs under Koch Trainer now stay within the custom set's learned-lesson prefix

### DIFF
--- a/Software/src/Version 6 and newer/MorsePreferences.cpp
+++ b/Software/src/Version 6 and newer/MorsePreferences.cpp
@@ -2020,8 +2020,7 @@ void MorsePreferences::writePreferences(const char* repository) {
 
      if (MorsePreferences::kochFilter != pref.getUChar("kochFilter")) {
         pref.putUChar("kochFilter", MorsePreferences::kochFilter);
-        if (!MorsePreferences::useCustomChars)                          // we update these only if we do not use a custom character set!
-          koch.setup();
+        koch.setup();
      }
 
      if (MorsePreferences::wlanSSID != pref.getString("wlanSSID")) {
@@ -2891,8 +2890,8 @@ void Koch::setup() {                                                // create th
   else
     setKochChars(MorsePreferences::pliste[posKochSeq].value);
   //// populate the array for abbreviations and words according to length and Koch filter
-  createWords(MorsePreferences::pliste[posWordLength].value, MorsePreferences::useCustomChars ? kochCharsLength+1 : MorsePreferences::kochFilter) ;  //
-  createAbbr(MorsePreferences::pliste[posAbbrevLength].value, MorsePreferences::useCustomChars ? kochCharsLength+1 : MorsePreferences::kochFilter);
+  createWords(MorsePreferences::pliste[posWordLength].value, MorsePreferences::kochFilter) ;  //
+  createAbbr(MorsePreferences::pliste[posAbbrevLength].value, MorsePreferences::kochFilter);
 
   String charSet = getCharSet();
   uint8_t probability = 0;


oe1wkl (commented):
Reviewed, built both variants, and traced the affected paths. **The fix is correct and I'd like to merge it.** `kochFilter` really is the lesson boundary for custom sets too — `getNewChar()`, `getRandomChar()` and `getCharSet()` all treat it that way; the word/abbrev tables were the only place that didn't.

### The bug was worse than the description says

`wordIsKoch()` returns exactly `kochCharsLength + 1` for a character that is **not in the set at all** (`MorsePreferences.cpp` ~2879), and the old bound was `kochCharsLength + 1` with a `<=` test. So under Custom Chars **every** word in `english_words.h` and every entry in `abbrev.h` passed the filter — not merely the ones past the current lesson. A digits-only custom set would still have served the full English word list. Worth capturing in the commit message for the record.

### Notes

**1. The Adaptive Random claim is overstated.** `MorseMenu.cpp:835` already calls `koch.setup()` on entry to Koch Echo Adaptive, so the probability table was rebuilt before it was ever used. The `writePreferences()` half of the change is still right — it removes the divergence — but the user-visible bug was Words/Abbrevs/Mixed only.

**2. The sibling guard survives.** `MorsePreferences.cpp:2109` still carries the same pattern:

```cpp
case posKochSeq:
      if (!MorsePreferences::useCustomChars)
        koch.setup();
```

Harmless today — when `useCustomChars` is true, `posKochSeq == 4`, and `handleKochSequence()` has already run `setup()` on every path that can change it (menu, snapshot recall, serial) — so it eliminates a redundant call rather than filtering anything. But removing one twin and leaving the other will make the next reader ask the same question again. I'll fold `posKochSeq` in with the `posAbbrevLength`/`posWordLength`/`posCarouselStart` cases in a follow-up.

**3. Adjacent pre-existing gap (out of scope here).** `PUT kochlesson/<n>` in `m32_v6.ino` (~4886) assigns `kochFilter` and returns — no `koch.setup()`, no `writePreferences()`. A lesson change over the serial protocol therefore leaves the word/abbrev tables built for the *previous* lesson (for built-in sequences as well as custom ones), and the value is lost on reboot. Same bug as this PR fixes for the on-device path; the config tool and the Accessibility Edition both drive lessons over the protocol. I'll take that one as a follow-up too.

### Duties checked

- **Builds:** `heltec_wifi_lora_32_V2` SUCCESS, `pocketwroom` SUCCESS — built here from the PR branch, not just taken on trust.
- **Manuals:** nothing owed. The V9 manual already documents the *fixed* behaviour — "at lesson *N*, only the first *N* characters of the set are used as the active training pool". The firmware was diverging from the docs; this restores agreement.
- **Accessibility (CLAUDE.md §8):** re-ran `extract_voice_strings.py`; `voice_strings.txt` diff is empty, so no clips owed.
- **Safety:** `koch.setup()` under custom chars calls `setCustomChars(customCharSet)`, which re-derives bounds from the in-memory string — it does **not** re-read `/player.txt`, so the new call in `writePreferences()` cannot clobber an active set the way the old prefs-exit bootstrap did.

### One behaviour heads-up

With a short or non-alphabetic custom set, "English Words" / "CW Abbrevs" under Koch Trainer will now often match zero entries and fall back to `getRandomChar(1)` — single random characters. That path is handled (no division by zero) and is identical to how built-in Koch behaves at low lessons, but anyone who had been getting real words out of a custom set will notice the change. Correct, just visible.

Thanks Christian — good catch, and the diagnosis of *why* it went wrong (the two call sites disagreeing about what `kochFilter` means under custom chars) is exactly right.

PR #210, opened 2026-08-30
Add a Practice Set frequency boost for Random Characters exercises

ckonecny:
## Summary
- New preference **Boost Practice** (Off / Moderate / Strong) for CW Generator and Echo Trainer's Random Characters mode, raising the odds that Practice Set characters (see the "Practice Set" picker) are drawn.
- Strictly intersected with the active **Random Groups** subset — a Practice Set character outside that subset (e.g. a digit while only Alpha is selected) is never boosted in; it has no effect, same as before this feature.
- **Also applies to the Koch Trainer's character generation** (both `Koch::getRandomChar()` — native sequence and Custom Chars — and `Koch::getAdaptiveChar()`), intersected with whatever the current Koch Lesson has unlocked. Implemented as a bounded reroll of Koch's own existing draw (redraw up to N times if the result isn't a Practice Set character) rather than a new weight table, so Koch's own weighting — favouring newer/adaptive characters — is left completely untouched. Willi's Koch pedagogy is unmodified either way.
- Scope is otherwise limited to character generation: Words, Abbreviations and Call Signs are unaffected (they're drawn as whole strings from fixed tables, so there's no clean way to "boost" one character into them without rejection-sampling and skewing their length/frequency distribution).
- User manual (EN + DE) updated with the new preference row (now covering both the Random Characters and Koch behaviour); HTML/PDF rebuilt.
- Accessibility voice strings extracted (`voice_strings.txt`/`voice_manifest.json`/`voice_clips.h` regenerated) for the 3 new strings ("Boost practice characters", "Moderate", "Strong"); re-running the extractor after the Koch extension produced no further diff (the preference/labels are reused, not new). The corresponding audio clips still need `generate_audio.sh` (Piper) run and both images reflashed before the `pocketwroom-accessibility` build ships with this — I didn't have the Piper voice model set up to render them myself.

## Test plan
- [x] Builds cleanly for `heltec_wifi_lora_32_V2` (Classic) and `pocketwroom` (Pocket)
- [x] Verified on device (M32 Pocket): CW Generator, Random Characters, Practice Set "y9", Boost = Strong → y/9 noticeably more frequent
- [x] Verified on device: Echo Trainer, same setup, same behaviour
- [x] CI green (PlatformIO CI, incl. manual-freshness check)
- [ ] `generate_audio.sh` for the 3 new accessibility voice clips, then reflash `pocketwroom-accessibility` with both `upload` and `uploadfs`


oe1wkl:
Thanks Christian — I've gone through this in detail. I checked out the branch and built all three shipping environments (`heltec_wifi_lora_32_V2`, `pocketwroom`, `pocketwroom-accessibility`): all clean, no warnings in the touched files. I also re-ran `extract_voice_strings.py` and got a zero diff against your committed tables, so the accessibility side is done correctly — only the clips themselves are outstanding, as you note.

The mechanics are right, and I deliberately checked the things that usually go wrong with a new preference:

* the three parallel arrays (`prefPos` enum / `pliste[]` / `prefName[]`) are positionally in sync;
* the mid-enum `prefPos` insertion is safe — NVS keys by name, snapshots are name-hash-tagged, and the protocol addresses configs by `parName`;
* `storedInSnapshot()` is opt-out, so the new preference is carried in snapshots automatically — which is what we want;
* `uint8_t weight[51]` is exactly the right bound (the widest `e - s` is 51) and the roll loop cannot overrun it;
* the boost block sits after both the `usePracticeChars` and `kochActive` early returns, so your scope claim holds.

Three things before this goes in.

### 1. `Koch::getAdaptiveChar()` can append a NUL

`char c = '\0'` and then `result += c;` unconditionally. On master, a draw that misses — the inner loop finishing without `randomOffset` going negative — was simply retried by the enclosing `while (result.length() < maxl)`; that retry *is* the miss handler. The new code appends `'\0'` into the generated group instead.

It is currently unreachable, because `Koch::setup()` zeroes `adaptiveProbabilities[]` beyond `charSet.length()`, so `getProbabilitySum()` cannot overshoot the range the inner loop walks. So this is hardening rather than a live bug — but it is one line, and it keeps the invariant explicit for whoever touches the probability table next. Please guard the append.

### 2. Two different mechanisms behind one label

The non-Koch path is an exact weighted draw (3× / 8×); the Koch path is bounded rejection sampling (3 / 8 attempts). They land in roughly the same place statistically, but they are not calibrated against each other, while the manual presents them as a single setting with two strengths. I am happy with the approach — your reasoning for not touching Koch's own weighting is right — but please add a sentence to the code comment saying the two levels mean different things on the two paths, so nobody later "fixes" one to match the other.

### 3. Discoverability: the boost is settable where the Practice Set is not

`posRandomBoost` is added to `generatorOptions`, `echoTrainerOptions`, `kochGenOptions` and `kochEchoOptions`, but `posPracticeChars` — the picker that defines the set being boosted — lives only in `allOptions`. So inside CW Generator's own preferences menu (black-knob double-click) a user can switch the boost on but can neither see nor set the Practice Set it acts on, and the preference will look broken.

Either add `posPracticeChars` to those same four arrays, or make the manual row say explicitly that the Practice Set is set from the top-level **Set Preferences**. I would prefer the first — it is where a user will look.

### Extending it into the Koch Trainer

That part is a product decision rather than a code review point, and I am fine with it: it defaults to Off, so nothing changes for anyone silently. Worth being aware that it lets a user re-weight the Koch sequence with a set they built for a different feature — which is exactly what someone drilling their problem characters would want, so I'll take it.

### Documentation

The EN and DE rows are well written, correctly cross-referenced and in the right table, and you rebuilt exactly the tracked set (the per-variant outputs are gitignored, so nothing is missing).

What is missing is a change-history entry in `Software/README.md` — this is user-visible and belongs in "New Features". Leave that to me, I'll write it in the same voice as the rest and run `sync_whatsnew.py`.

### Merging

The branch is 8 commits behind master and conflicts only on the generated `manual_*.html` / `*.pdf`. Those get regenerated, not hand-merged. If #211 lands as well, the two additionally collide on `voice_clips.h` and `voice_manifest.json` — same treatment, take either side and re-run the extractor afterwards. I'll do a single `generate_audio.sh` run covering the clips owed by both PRs and reflash `pocketwroom-accessibility` with `-t upload` and `-t uploadfs`, so you do not need the Piper setup for this.

ckonecny:
Thanks for the thorough review — addressed all three:

1. **NUL append in `getAdaptiveChar()`**: the reroll candidate now defaults to `charSet.charAt(0)` instead of `'\0'`, so even if the `adaptiveProbabilities[]`/`charSet` invariant is ever broken later, the generated group can no longer be corrupted with a NUL byte. Agreed it's currently unreachable — this is purely hardening.
2. **Two mechanisms, one label**: added a comment in both `getRandomChars()` (m32_v6.ino) and `practiceBoostAttempts()` (MorsePreferences.cpp) stating explicitly that the weighted draw and the bounded rejection sampling are separate, independently-tuned mechanisms and are not meant to be numerically identical.
3. **Discoverability**: added `posPracticeChars` to `generatorOptions`, `echoTrainerOptions`, `kochGenOptions` and `kochEchoOptions` (your preferred fix) — the Practice Set picker is now reachable from every menu where Boost Practice itself is.

Pushed as 374770c. Left the `Software/README.md` change-history entry and the merge/rebase to you as noted.

oe1wkl:
Checked 374770c — all three addressed, and I verified each rather than taking the summary on trust: `posPracticeChars` is in all four option arrays and sensibly grouped, the "these are two different mechanisms, do not calibrate them against each other" note is in both places, and `char c = charSet.charAt(0)` replaces the `'\0'` with a comment that explains why the placeholder matters at all. Builds clean for all three shipping environments here.

One leftover only, and I will fix it when I merge: `displayKeyerPreferencesMenu()` still says `// reached via allOptions, nestled among the "Set Preferences:" items` next to the `posPracticeChars` case. It is reached from five arrays now. Behaviour is right, only the comment is stale.

The change-history entry in `Software/README.md`, the `sync_whatsnew.py` sync and the manual rebuild are mine, as is the voice-clip run — I will do those together with #211 so the two only cost one `generate_audio.sh` and one reflash of the Accessibility Edition.

Thanks — this one is ready as far as the code goes.

PR #211, opened 2026-08-30
Koch Trainer: "Preview Char" - pick any lesson character to repeat

ckonecny:
## Summary
- New Koch Trainer menu entry, sibling to "Learn New Chr": a picker (in lesson order, across the whole course) to choose any character and hear it repeated - handy for revisiting an earlier character or previewing one further ahead, without waiting for the lesson to get there.
- Implemented as a new `GEN_TYPE` value (`KOCH_PREVIEW`) riding the existing `echoTrainer`/Learn-New-Char machinery (same repeat behavior, same speed control via the encoder) rather than a new mode - reuses `Koch::getKochChar()` and the `selectFilePart()`-style picker pattern already used for the file-player chapter picker.
- Triple-clicking the FN button during playback reopens the picker to switch characters without leaving the mode (that click count was previously unassigned).
- Along the way, fixed the picker's own input loop to also honor FN double-click (brightness) and mode-button double-click (preferences menu) - those silently did nothing while any `selectFilePart()`-style picker screen was open, since it blocks in its own loop outside the main `loop()`. Only fixed for this new Koch Preview screen, not backported to the older File Player chapter picker.
- The picker's visible-line count now follows `NoOfVisibleLines` instead of a hardcoded 3, so the M32 Pocket shows more of the list (4/5 lines depending on Font Size) instead of always just 3.
- EN/DE user manuals updated with a new "Koch: Preview Char" section.

## Known follow-up
Voice-clip tables (`voice_strings.txt`, `voice_manifest.json`, `voice_clips.h`) were regenerated via `extract_voice_strings.py` for the new "Preview Char" menu label, but the actual audio clip has **not** been rendered (no local Piper/TTS model setup in the environment this was built in). Before refreshing the `pocketwroom-accessibility` build, run:
```
cd Software/tools/audio-accessibility
./generate_audio.sh
```
then reflash both `-t upload` and `-t uploadfs` for that environment. Until then, the Accessibility Edition will report a voice-pack stamp mismatch at boot (by design - not a silent failure) rather than staying silent on this one entry.

## Test plan
- [x] Builds cleanly for both `heltec_wifi_lora_32_V2` and `pocketwroom`
- [x] Flashed to a physical M32 Pocket and tested interactively: picker navigation, character playback, live speed change, triple-FN back-to-picker, double-click brightness/preferences from within the picker, long-press exit
- [ ] Not tested on Classic M32 hardware (OLED) - only build-verified
- [ ] Accessibility Edition voice clip for "Preview Char" not yet rendered (see Known follow-up)


oe1wkl:
Thanks Christian — a genuinely useful feature, and I like that you rode the existing `echoTrainer` machinery instead of inventing a mode. I checked out the branch and built all three shipping environments (`heltec_wifi_lora_32_V2`, `pocketwroom`, `pocketwroom-accessibility`): all clean. Re-running `extract_voice_strings.py` gives a zero diff against your committed tables, and the `spoken_overrides.tsv` entry is exactly right — that is the accessibility duty done properly.

The `NoOfVisibleLines` handling in the picker is correct and a real improvement.

There are three things I need changed before this goes in, and a handful of smaller ones.

---

## 1. `_kochPreview` must be appended at the end of the enum, not inserted

This is the important one. Inserting it at position 19 shifts every `menuNo` from `_kochGen` upward, and `menuPtr` is persisted as a raw index:

* `pref.putUChar("lastExecuted", menuPtr)` in `MorsePreferences::writeLastExecuted()` — so **Quick Start launches the wrong mode after the upgrade** for anyone whose last mode was at index ≥ 19;
* every stored **snapshot** carries `lastExecuted` as a raw index too, and those stay wrong until the snapshot is re-stored — a silent, permanent corruption of the user's saved presets;
* `GET menus` reports the numeric index to protocol clients (`jsonMenu(..., (unsigned int) ptr, ...)`).

The convention in this codebase is the opposite, and it is written down right where you would need it — `_genPractice` carries the comment *"appended at enum's end; spliced into the gen ring via the `_genPlayer`/`_genRand` edits above"*, and the games entries (`_trailblazer`, `_foxHunt`, `_memoryChain`) do the same.

The fix is mechanical: move `_kochPreview` to the end of the `menuNo` enum, and its `menuText[]` and `menuNav[]` rows to the matching end position. Your nav wiring is already correct and does not change — display order comes from `menuNav`, not from the enum.

## 2. Prosigns are rendered raw in the picker

`String(idx + 1) + ": " + koch.getKochChar(idx)` never goes through `cleanUpProSigns()`. In the native sequence that hits six entries — positions 43, 45, 46, 47, 48, 49 draw as bare `K S N A E B`, duplicating the plain letters already listed at positions 2, 4, 14, 6, 16, 31. The picker therefore shows both "2: k" and "43: K", differing only in case.

This is exactly what `practiceGlyph()` in `MorsePreferences.cpp` was written to prevent — its comment cites hard rule 4 by name and makes the same argument. Please clean the character for display, and keep the raw one for the accessibility announcement, the way `announceValue()` documents for `posKochFilter`.

## 3. Please replace the triple-click with a third encoder role

I do not want to ship a triple-click. Two reasons:

**It would be the first one.** I grepped all 235 `.clicks` sites in the firmware — no triple-click handler exists anywhere today. This is not merely undocumented in `UX_CONVENTIONS.md`, it is a gesture the device has never asked of a user.

**Its failure mode is wrong.** `multiclickTime` is 220 ms and `ClickButton` only reports the count once the window closes, so a third press that lands late does not fail neutrally — it becomes a double-click and **steps the display brightness**. "Give me the picker" and "dim the screen" should not be the same gesture with a 220 ms tolerance.

There is also a correctness problem with the guard as written. The `case 3:` block lives in the **global** `loop()` volume-button switch, which runs in every mode, and it tests only `generatorMode == KOCH_PREVIEW`. `generatorMode` is sticky — `MorseMenu.cpp` already resets it defensively in the transceiver cases (`generatorMode = RANDOMS; // to reset potential KOCH_LEARN`) for precisely this reason, but the Keyer and the Decoder do not. So: Preview Char → long-press out → CW Keyer → red triple-click opens the Koch picker on top of the Keyer and sets echo-trainer state underneath it.

### What I would like instead

Give the encoder a third role in this mode, selected by the red button's existing role cycle. `UX_CONVENTIONS.md` already frames the encoder as owning a *role* — "speed by default, or volume / scroll-back / keyer-memory after a red-button gesture selects that role" — so this extends a documented mechanism rather than adding a gesture beside it. In Preview Char only, the red single-click would cycle **speed → volume → character → speed**, and turning the encoder then steps through the course live, in place: no picker overlay, no mode restart.

I think this is also the better feature. You are auditioning characters, so stepping through them while they play is the natural motion — the picker is the thing you use to *arrive* somewhere, the encoder is the thing you use to *browse*.

Sketch of what it needs:

* append a value to `encoderMode` in `morsedefs.h` (`previewCharMode` or similar). Safe to append: `encoderState` is runtime-only, never persisted, and `menu_()` already resets it to `speedSettingMode` on every menu entry — so unlike the triple-click, the new role cannot leak into another mode;
* extend the `volButton` `case 1:` cycle to three stops, gated on `morseState == echoTrainer && generatorMode == KOCH_PREVIEW` (both conditions — see the sticky-`generatorMode` problem above);
* add a `case previewCharMode:` to the encoder dispatch switch alongside `case memSelMode:`, stepping `previewCharIndex` within `[0, kochMaximum)` and restarting playback on the new character;
* show which role the encoder owns. The established idiom is the bold/normal flip in the status line — `displayCWspeed()` and `MorseOutput::displayVolume()` both key off `encoderState`. The OLED is the tighter target here, so the character indication needs a column budget that fits three visible lines.

Keep the picker at mode entry — 51 characters is too many to spin through blind, and the numbered lesson-order list is genuinely useful for jumping. It is the *in-mode re-entry* gesture I want replaced, and with it the whole `case 3:` block.

I'll add the new role to `UX_CONVENTIONS.md` once the shape is settled.

---

## Smaller items

**Stale option set in the picker.** `setCurrentOptions(kochEchoOptions, …)` is called *after* `selectKochPreviewChar()` in the `_kochPreview` case, so the preferences double-click you added inside the picker opens whatever option set the previous mode left behind — `allOptions` at boot, `loraTrxOptions` after a LoRa session. Move that line above the picker call.

**Stale bounds after that.** `count` is read once from `kochMaximum` at entry, but the in-picker preferences let the user change Koch Sequence or Custom Chars, which changes `kochMaximum`. Neither `count` nor `selected` is re-read, so an out-of-range index reaches `koch.getKochChar()` and yields an empty `clearText`. Re-read and re-clamp after `setupPreferences()` returns.

**The picker is silent on the Accessibility Edition.** The `jsonCreate("message", …)` line serves external clients, but the on-device edition says nothing — and this is a screen a blind user has to navigate. The machinery already exists and fits exactly: `announceValue()`'s `posKochFilter` case does `a11ySay(n)` / `a11ySay("of")` / `a11ySay(total)` / `MorseVoice::announceMoreChar(char)` and produces "21 of 51, Yankee". The same call with `idx + 1`, `count` and `koch.getKochChar(idx)` gives the picker a voice. (This is why the raw character matters in item 2 — `announceMoreChar` needs it before `cleanUpProSigns`.)

**`kochFilter == 0` edge.** `selected = kochFilter - 1` leaves `selected` at `-1`, and it is only clamped when the encoder moves — an immediate click then returns `-1`, which the caller reads as "cancelled". Cheap to clamp at the point of assignment.

**Remote execution.** `_kochPreview` passes `isRemotelyExecutable()` because its `naviDown` is 0, but unlike `selectFilePart()` the picker has no remote-selection path (`remoteFilePartSelect` and friends). A protocol client executing it strands the device in a picker only someone standing at the device can dismiss. Either wire the remote path or exclude `_kochPreview` in `isRemotelyExecutable()`.

## Documentation

The EN and DE sections are well written and in the TOC, and you rebuilt exactly the tracked set. Two additions once the control changes: the manual will need the new encoder role described instead of the triple-click, and it should mention the in-picker double-click gestures you added (brightness, preferences).

A change-history entry in `Software/README.md` is missing — leave that to me, I'll write it in the same voice as the rest and run `sync_whatsnew.py`.

## Merging

The branch is 8 commits behind master and conflicts only on the generated `manual_*.html` / `*.pdf`, which get regenerated rather than hand-merged. If #210 lands as well, the two additionally collide on `voice_clips.h` and `voice_manifest.json` — same treatment. I'll do a single `generate_audio.sh` run covering the clips owed by both PRs and reflash `pocketwroom-accessibility` with `-t upload` and `-t uploadfs`, so you do not need the Piper setup for this.

ckonecny:
While testing this branch I ran into three pre-existing display/state bugs, unrelated to Preview Char itself — flagging them here rather than fixing them in this PR to keep it scoped.

**1. Status line corrupts when changing speed/volume while paused.** Pause via ENCODER-knob click (shows "Continue with paddle"), then turn the ENCODER to change speed — the line becomes garbled, e.g. `Con(15)60wWpM paddle`. Root cause: `printOnStatusLine()` only clears the pixel width of what it draws, not the full line, so `displayCWspeed()`/`MorseOutput::displayVolume()` patch narrow fields on top of the wider "Continue with paddle" message instead of replacing it, leaving fragments behind. (I hit the same pattern in Preview Char's own new code and fixed it there with a full `updateTopLine()` repaint instead of narrow patches — same fix would apply here.)

**2. Re-selecting "Custom Chars" in Koch Sequence empties it.** Switch Koch Sequence away from Custom Chars and back on-device — the custom set comes back empty. `MorsePreferences::handleKochSequence()` is called with `forceReload=true` on every on-device (re-)selection, which discards the cached `customCharSet` and always re-reads `/player.txt` via `getCustomChars()` from scratch; if that re-read comes back empty, the set is lost (the code's own fallback — reverting Koch Sequence to M32 — doesn't seem to be what's observed, worth a closer look at why the re-read comes back empty the second time).

**3. "No custom set" leaves a leftover character.** Scrolling through Koch Sequence options, landing on "Custom Chars" with none available briefly shows "No custom set" (`MorseOutput::printOnScroll`), then the next option (e.g. "M32") is drawn — but the leading "N" of "No custom set" stays on screen. Same root cause as #1: `printOnScroll()` only clears the width of the new (shorter) string, not the wider one it's replacing.

Happy to fix any/all of these in a separate PR if you'd like — just say the word.

ckonecny:
Pushed e7bcffa, addressing all the feedback:

**The three must-fix items:**
1. `_kochPreview` moved to the end of the `menuNo` enum; `menuText[]`/`menuNav[]` rows moved to match. The Koch submenu sibling wiring itself is unchanged (display order comes from `menuNav`'s links, not enum position). Also caught a related miss while testing: `_kochPreview` needed adding to the Quick Start "remember this leaf" check in `MorseMenu.cpp` too — otherwise entering Preview Char never updated `lastExecuted` at all. Found this via on-device testing (Quick Start kept resuming into a stale earlier mode) and fixed it.
2. Prosigns now go through `cleanUpProSigns()` for display in the picker, raw character kept for the accessibility announcement.
3. Triple-click replaced with the third encoder role you sketched (speed → volume → character → speed, Preview Char only), with the bold-glyph status-line indicator at column 1. While building this I hit two follow-on bugs, both fixed: `changePreviewChar()` wasn't clearing `CWword` itself (only the read position), so it kept replaying the character being left instead of ever reaching the new one; and cycling the encoder role — or stepping characters — while paused ("Continue with paddle") corrupted the status line, same root cause as bug #1 in my comment above, fixed here the same way (full `updateTopLine()` repaint instead of narrow field patches).

**The five smaller items:** all done — `setCurrentOptions()` moved ahead of the picker call, count/selection re-read and re-clamped after the in-picker preferences double-click, the picker now voices itself on the Accessibility Edition, the `kochFilter == 0` edge is clamped at assignment, and `_kochPreview` is excluded from `isRemotelyExecutable()`.

**Tested on a physical M32 Pocket**, on top of building all three of your environments clean: picker navigation and confirm/cancel, prosign display at lesson positions 43/45-49 vs. the plain-letter duplicates at 2/4/6/14/16/31, the encoder role cycle and live character browsing, in-picker double-click (brightness, preferences — including changing Koch Sequence/Custom Chars mid-picker), Quick Start resuming correctly into Preview Char after a fresh selection, and long-press exit from both the picker and playback.

EN/DE manuals updated for the encoder role and the picker's double-click gestures.

oe1wkl:
Thanks Christian — I built e7bcffa for all three shipping environments (`heltec_wifi_lora_32_V2`, `pocketwroom`, `pocketwroom-accessibility`): all clean.

I should say plainly that I had implemented the same three must-fix items myself in parallel while you were working, on a local branch, and tested that on classic hardware. So this is a comparison of two independent goes at the same thing rather than a plain review — and yours is the better base. Two things in particular you caught that I did not:

* **`_kochPreview` in the Quick Start leaf check.** Moving it past `_wifi` makes `menuPtr < _wifi` false, so `lastExecuted` never updates. My branch has exactly that bug — I read that line while checking for range comparisons and judged it safe, which it wasn't. You also improved the warning comment there instead of dropping it, which is the right instinct: that comment is the only thing standing between the next person and the same trap.
* **The paused status line.** Your full `updateTopLine()` repaint on every role transition is more robust than the targeted repaint I had, and it fixes a case I had not thought about at all.

## Three bugs in the new code

All three are in the accessibility and status-line work, and I don't think any of them was catchable with the setup you tested on: `pocketwroom` is the standard build, where every `MorseVoice::*` call compiles to a no-op. Together, the first two mean the picker's voice support is currently inert.

**1. `cleanUpProSigns()` modifies its argument.** It takes `String&` and ends with `input = buf; return input;`. So in

```cpp
String rawChar   = koch.getKochChar(idx);
String cleanChar = cleanUpProSigns(rawChar);   // rawChar is now "<ka>" too
...
if (idx == selected) { curLine = line; curRawChar = rawChar; }
```

`curRawChar` carries the expanded tag, and `announceMoreChar()` gets `<ka>` where it needs `K`. The comment immediately above the call states the requirement correctly ("character (pre-cleanUpProSigns) so prosign clips match") — it's only the aliasing that defeats it. It breaks precisely the six prosign positions and leaves the plain letters working, so it is easy to miss. Simplest fix is to take a fresh copy for the announcement, or clean into a separate variable.

**2. No `MorseVoice::tick()` in the picker loop.** `selectKochPreviewChar()` blocks in its own loop outside `loop()`, and `tick()` is what actually starts and advances the clips — it is why `setupPreferences()` and the menu loop both call it. Without it the announcements are queued and never played.

**3. `displayPreviewChar()` is guarded only on `generatorMode`,** and `updateTopLine()` calls it unconditionally. `generatorMode` is sticky across modes — the same trap you guarded against in the role cycle, with the reasoning spelled out in the comment there. It is reachable without leaving the Keyer: use Preview Char, long-press out, enter CW Keyer, then select a keyer memory (that path calls `updateTopLine()`), and a stray character appears at column 1 and stays. The role-cycle condition — `morseState == echoTrainer && generatorMode == KOCH_PREVIEW` — is what this guard wants too.

**One nit:** the total is announced on every selection change. `announceValue()`'s `posKochFilter` case deliberately speaks it on entry only, and says why: "repeating it on all 51 detents makes the list far slower to scan by ear."

## The status-line indicator — one open question

This is the one place where I would like to keep my version, and I would rather put the reasoning to you than just overrule it.

```
        cols:  0  1  2  3  4  5  6  7  8  9  10 11 12 │ 13 … 17
 yours:           K [i]  (  1  2  )  2  0     W  p  M │ meter │ logo
 mine:              [i]  4  3  :  <  k  a  >  C  h  r │ meter │ logo
                         ^^^^^^^^^^^^^^^^^^^ bold
```

Yours is the less disruptive of the two — column 1 really is free in echoTrainer, and it keeps both WpM readouts. But it puts a bare prosign code on the screen: a lone `K` for `<sk>`, distinguished from the letter `k` by case alone. That is the thing hard rule 4 exists to prevent, and it is the same argument your own picker fix rests on. Your note says a lone glyph cannot be confused the way a numbered list can — I don't think that holds, because the browsing user has nothing on screen to compare it against, whereas in the picker both spellings are visible at once and the ambiguity is at least apparent.

The cost of my version is both WpM numbers while the encoder owns the character. Neither is adjustable in that role and one click brings them back, so I think that trade is worth it — and it buys room for the position number as well, which makes stepping to a known lesson much easier. It fits with 2 px to spare before the volume meter, the same margin `WpM` already has, and I have had it on a classic M32: the fonts are strictly monospaced at 7 px in both weights, so the column arithmetic is exact.

Worth flagging that your layout has not been on an OLED — the Pocket has considerably more room, and 18 columns is the tighter target.

## Next

I will port the three fixes onto your branch myself rather than send you round again, and settle the status line while I am in there. Nothing needed from you on this one.

## Your three pre-existing bugs

Yes please, in a separate PR — good finds, and right call keeping them out of this one.

**Start with #2 (Custom Chars emptied on re-selection).** That one loses the user's data, and it is in `handleKochSequence(forceReload)`, which is recent — it came in with #200. The other two share a root cause, so decide first whether the fix belongs in `MorseOutput` itself (clearing the full field rather than the drawn width) or at the call sites; the library-level fix is tidier but costs a full-width clear on every status write, which on the OLED means flicker. I would take a look at that trade before writing either.

PR #212, opened 2026-08-31
Fix three pre-existing display/state bugs found while testing PR #211

ckonecny:
## Summary

Three pre-existing bugs, unrelated to the Preview Char feature, found and flagged while testing PR #211 (see [the discussion there](https://github.com/oe1wkl/Morserino-32/pull/211#issuecomment-5477423449)). Willi asked for these to go into a separate PR, prioritizing the Custom Chars data-loss bug first.

## 1. Koch Sequence "Custom Chars" comes back empty on re-selection

`getWord()` tracked "did we just wrap past EOF" in a function-local `static`, shared by three unrelated readers of `player.txt`: the running File Player, `skipWords()` (fast-forwards to a saved resume position on File Player entry), and `getCustomChars()` (full-file scan for the Custom Chars set). If either of the first two happened to wrap exactly where they stopped reading, the flag was left `true`, and the next `getCustomChars()` call saw it on its very first read and bailed out with an empty set - even though the file was fine.

Fix: hoisted the flag (`wordReaderAtEof`) to file scope so `getCustomChars()` can reset it before its own scan, making that scan self-contained regardless of what any other reader left behind.

Tested: switched Koch Sequence away from Custom Chars and back on-device with a populated `player.txt` - custom set stays intact.

## 2. Status line corrupts when changing speed/volume while paused

Pause via the ENCODER-knob click ("Continue with paddle"), then turn the ENCODER to change speed, or click FN to toggle speed/volume - the line became garbled, e.g. `Con(15)60wWpM paddle`.

Root cause: `printOnStatusLine()` only clears the pixel width of what it draws, not the full line, so a narrow field update (WPM digits, volume bar) on top of that wider paused message left fragments of it behind.

First pass added a per-line "widest full-line text still potentially visible" tracker (`statusLineCacheWidth`) inside `printOnStatusLine()` so a narrow write clears at least that far right too. That fixed the fields the narrow write actually touches, but not columns to the *left* of it - turning the encoder to change speed only touches columns 3/7/10, so "Continue with paddle"'s leading "Con" (columns 0-2) was never reached and stayed put. Fixed properly by adding `MorseOutput::statusLineHasForeignContent()` and checking it at the three places a narrow update can happen right after a full-line message: `changeSpeed()`, `changeVolume()`, and the FN-click speed/volume role toggle - each now does a full `updateTopLine()` repaint instead of its normal narrow update when that's the case, falling back to the narrow (fast, no-flicker) path the rest of the time.

Tested: paused, then changed speed by turning the encoder, and separately by clicking FN to toggle to volume - both stay clean. Also confirmed normal (non-paused) encoder operation is unchanged - no added flicker or lag.

## 3. "No custom set" leaves a leftover character

Scrolling through Koch Sequence options, landing on "Custom Chars" with none available briefly shows "No custom set" (`MorseOutput::printOnScroll`, xpos 0), then the next option's value is drawn at xpos 1 - the leading "N" of "No custom set" was never touched and stayed on screen.

Tried the same generic fix as #2 in `printOnScroll()` first, but reverted it: `printOnScroll()` has a legitimate pattern - `MorsePreferences::setupPreferences()` draws a value at xpos 1 and then a narrow single-space "cleanup" write at xpos 0 on the *same* row right after - that a generic "clear the widest-ever-written extent" rule misreads as stale content needing cleared, wiping the value that was just drawn (caught via on-device testing before this ever reached a PR). Fixed at the one call site instead: an explicit `MorseOutput::clearLine()` right after the "No custom set" message, rather than in the shared library function.

Tested: scrolled to Custom Chars with an empty `player.txt` ("No custom set" shown), then to the next option - no leftover character.

## Test plan
- [x] Builds cleanly for all three shipping environments: `heltec_wifi_lora_32_V2`, `pocketwroom`, `pocketwroom-accessibility`
- [x] All three fixes verified on a physical M32 Pocket, including the regression found and fixed along the way (see commit history)
- [ ] Not tested on Classic M32 hardware (OLED) - only build-verified

### DIFF
--- a/Software/src/Version 6 and newer/MorseOutput.cpp
+++ b/Software/src/Version 6 and newer/MorseOutput.cpp
@@ -747,6 +747,14 @@ namespace {
     bool   statusClearPending = false;   // clearStatusLine was called, not yet flushed
     String statusLineCache;              // text last drawn via printOnStatusLine at xpos=0
     bool   statusLineStrong  = false;    // strong flag for that cached text
+    // Pixel width of the widest xpos==0 (full-line) text still potentially visible on
+    // screen. Unlike statusLineCache (invalidated by the very next narrow write, since the
+    // visible text is then no longer a single known string), this is deliberately NOT reset
+    // by a narrow write - only by something that actually wipes the line/screen. A narrow
+    // write (WPM digits, keyer symbol, ...) must clear at least this far right, or it leaves
+    // fragments of whatever wider full-line message (e.g. "Continue with paddle") is still
+    // sitting past its own narrow width - see printOnStatusLine().
+    uint16_t statusLineCacheWidth = 0;
 
     // Paint the full-width white background, respecting the battery icon's
     // right-side reserved area. Extracted so clearStatusLine() and the
@@ -785,6 +793,7 @@ void MorseOutput::initDisplay()
   // printOnStatusLine actually repaints rather than short-circuiting.
   statusClearPending = true;
   statusLineCache    = "";
+  statusLineCacheWidth = 0;
 }
 
 
@@ -818,6 +827,7 @@ void MorseOutput::clearDisplay() {
     // full background by treating it as if clearStatusLine was just called.
     statusClearPending = true;
     statusLineCache    = "";
+    statusLineCacheWidth = 0;
 #ifdef CONFIG_MCP73871
     batteryDisplayDirty = true;
     batteryIconVisible = false;
@@ -1861,6 +1871,7 @@ void MorseOutput::printOnStatusLine(boolean strong, uint8_t xpos, const String& 
   if (statusClearPending) {
     paintStatusBackground();
     statusClearPending = false;
+    statusLineCacheWidth = 0;   // background is genuinely blank now
   }
 
   if (strong)
@@ -1869,19 +1880,32 @@ void MorseOutput::printOnStatusLine(boolean strong, uint8_t xpos, const String& 
     display.setFont(DialogInput_plain_12);
   display.setTextAlignment(TEXT_ALIGN_LEFT);
   uint16_t w = stringWidth(string);
+  uint16_t x = xpos * display.getStringWidth("A");
+  // Clear at least as far right as the last full-line write reached (statusLineCacheWidth):
+  // a narrower write here (e.g. WPM digits) must not leave fragments of a wider message
+  // (e.g. "Continue with paddle") sticking out past its own width. Only extends the clear
+  // when that old content actually reaches into or past this xpos - never shrinks it.
+  uint16_t clearW = w;
+  if (statusLineCacheWidth > x) {
+      uint16_t neededW = statusLineCacheWidth - x;
+      if (neededW > clearW) clearW = neededW;
+  }
   display.setColor(WHITE);
-  display.fillRect(xpos * display.getStringWidth("A"), 0 , w, SCROLL_TOP);
+  display.fillRect(x, 0 , clearW, SCROLL_TOP);
   display.setColor(BLACK);
-  display.drawString(xpos * display.getStringWidth("A"), 0, string);
+  display.drawString(x, 0, string);
   display.setColor(WHITE);
   display.display();
 
-  // Only full-line writes (xpos==0) become the cached value; partial
-  // updates (WPM digits, keyer mode, etc.) invalidate the cache since
-  // the visible text is no longer a single known string.
+  // Only full-line writes (xpos==0) become the cached value; partial updates (WPM digits,
+  // keyer mode, ...) invalidate statusLineCache since the visible text is no longer a single
+  // known string - but deliberately leave statusLineCacheWidth alone: it keeps guarding
+  // narrow writes against the wider text until a new full-line write replaces it, or
+  // clearStatusLine()/clearDisplay() actually wipes the line/screen.
   if (xpos == 0) {
     statusLineCache  = string;
     statusLineStrong = strong;
+    statusLineCacheWidth = w;
   } else {
     statusLineCache = "";
   }
--- a/Software/src/Version 6 and newer/MorseOutput.cpp
+++ b/Software/src/Version 6 and newer/MorseOutput.cpp
@@ -1916,7 +1916,7 @@ void MorseOutput::printOnStatusLine(boolean strong, uint8_t xpos, const String& 
   #endif
 }
 
-boolean MorseOutput::statusLineHasForeignContent() {
+boolean MorseOutput::statusLineHasFullLineMessage() {
   return statusLineCacheWidth > 0;
 }
 
--- a/Software/src/Version 6 and newer/MorseOutput.cpp
+++ b/Software/src/Version 6 and newer/MorseOutput.cpp
@@ -1916,6 +1916,10 @@ void MorseOutput::printOnStatusLine(boolean strong, uint8_t xpos, const String& 
   #endif
 }
 
+boolean MorseOutput::statusLineHasForeignContent() {
+  return statusLineCacheWidth > 0;
+}
+
 
 
 void MorseOutput::clearStatusLine() {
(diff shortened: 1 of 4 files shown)


oe1wkl:
Thanks Christian — and thanks for rebasing onto current master, including the `cleanUpProSigns()` change I pushed only a little before you opened this. GitHub still shows the PR as conflicting against 920839f; that is stale metadata, the branch is actually current. Your delta is two commits over four source files, with no manual artefacts, which is exactly right.

I have built all three shipping environments here (`heltec_wifi_lora_32_V2`, `pocketwroom`, `pocketwroom-accessibility`) — all clean — and put it on a classic M32 to cover the OLED gap in your test plan. I will report back on that.

## The thing I most wanted to check

The paused-line fix lands in `changeSpeed()` and `changeVolume()`, not in `changeSpeedValue()` / `changeVolumeValue()`. That matters more than it looks: the `*Value()` cores exist precisely because every game calls them to get the clamp, timings and protocol *without* the classic top-bar draw. Had `updateTopLine()` gone into the cores, it would have painted the classic status line over a running game's HUD on the Pocket.

Worth flagging because the diff reads alarmingly — the hunk header says `@@ void changeSpeedValue( int t) {` — but that is only the enclosing-function label; the modified lines are in the wrapper below it. Anyone reviewing this quickly could reasonably conclude the opposite. I verified it against every `changeSpeedValue`/`changeVolumeValue` call site.

I also like that the mechanism is self-limiting: the first detent after a full-line message costs a full repaint, `updateTopLine()` then clears `statusLineCacheWidth`, and every detent after that takes the fast narrow path. No sustained flicker.

And the best part of the PR is fix 3 — you tried the generic `printOnScroll()` version, found it broke `setupPreferences()`'s legitimate "value at xpos 1, then a single-space cleanup at xpos 0 on the same row" pattern, and reverted to a call-site fix. That is the right instinct about a shared drawing primitive, and the reasoning being in the PR body is worth as much as the code.

## One thing to finish

`getCustomChars()` resets `wordReaderAtEof` before its scan, but leaves it `true` afterwards — the scan runs to EOF, so the flag is set on the way out. The next reader of `player.txt` then sees it on its first call and swallows a word. That is the same bug you fixed, in the mirror direction: entering File Player straight after a Custom Chars re-selection should lose its first word.

It is pre-existing rather than introduced here, and much less serious than an empty character set — but it is the same one-line remedy, and doing both makes the function genuinely self-contained rather than self-contained in one direction. Please reset it again after the reopen at the end.

## One decision for me, not for you

After the repaint, "Continue with paddle" is gone, and nothing else on screen indicates the mode is paused — `genIsActive` has no other visual. So changing speed while paused now silently removes the only cue that you are paused. That is plainly better than the garbled line it replaces, and while the operator is adjusting speed the speed is the thing they want to see. But it is a visible behaviour change to a classic mode, so it is mine to settle rather than something to spring on people in a release.

I will decide it when I have had the classic in front of me for a while. Either the current behaviour stands, or the prompt is restored once the encoder goes quiet. Nothing for you to do unless I come back and ask.

## One naming point

`statusLineHasForeignContent()` returns true after *any* xpos==0 write, not only a foreign one. It happens to be accurate today, because in the running modes the only full-line writers are the paused and stop messages — but the name promises a judgement the function does not make, and the next full-line writer added anywhere will quietly inherit the "foreign" verdict. Something like `statusLineHasFullLineMessage()` would say what it actually measures and would age better. Your call; I will not hold the PR for it.

Once the `getCustomChars()` reset is in, I will merge this.

oe1wkl:
Tested now on both a classic M32 (the OLED gap in your test plan) and an M32 Pocket. All three fixes do what they claim. Four things came out of it — two are defects that need fixing, two are decisions I said I would make and have now made.

## 1. `getCustomChars()` destroys the File Player's saved position — and persists it

This is the important one, and it is bigger than the leftover EOF flag I mentioned before.

Entering File Player straight after a Custom Chars re-selection starts at the first word of `player.txt`, not where you left off. `getCustomChars()` sets `MorsePreferences::fileWordPointer = 0` on entry and never restores it, and `writeWordPointer()` then writes that zero to NVS:

```cpp
if ((MorsePreferences::fileWordPointer != pref.getUInt("fileWordPtr")))
    pref.putUInt("fileWordPtr", MorsePreferences::fileWordPointer);
```

So it is not just this session — the saved position is gone for good. For anyone working through a long text that is worse than the empty character set you fixed, and it is the same root shape: `getCustomChars()` borrows shared reader state and does not put it back.

Please make the function properly self-contained: save `fileWordPointer` on entry and restore it on the way out, alongside the `wordReaderAtEof` reset I asked for before (which is still needed in its own right — reset it again after the final reopen, not only before the scan).

Incidentally this corrects my earlier guess. I predicted the leftover EOF flag would show up as a swallowed first word; the actual symptom is the reset pointer, which is a different mechanism in the same function. Both are real.

## 2. "No custom set" — the fix covers the wrong half

You clear the line *after* the 700 ms message. The leftover character I reported is visible *during* it: with **LICW Carousel** previously selected, its trailing `l` sits there next to the message.

The arithmetic: option values are drawn at **xpos 1** (all five `printOnScroll(2, REGULAR, 1, valueLine)` sites), "No custom set" at **xpos 0**. The message is 13 characters, columns 0–12; "LICW Carousel" at xpos 1 occupies columns 1–13. Column 13 is never touched. Your `clearLine(2)` tidies up before the *next* option is drawn, which is also worth doing — but the message itself needs a clear before it, or it will keep inheriting the tail of whatever longer value it covers.

## 3. Decision: keep the revert to M32, but say so

I had asked myself whether the automatic jump to M32 should stay. It should. `handleKochSequence()` does it deliberately, and the comment there gives the reason — leaving the menu showing "Custom Chars" while the engine runs a different sequence is the worse state, and self-healing that is the point.

What made it feel like a glitch is that 700 ms is too short to read, and the jump is unexplained. So: lengthen the message and let it say where it landed — "No custom set" followed by something naming M32 — so it reads as a decision rather than the display losing its place. Take whatever timing feels right when you have it in front of you; 700 ms is definitely not it.

## 4. Decision: the paused prompt should come back

On your open question from the PR body — after the repaint, "Continue with paddle" is gone and nothing else says the mode is paused. Testing it did not bother me much, but I knew what to expect, which is exactly the wrong basis for judging it.

Please have the prompt reappear once the encoder has been quiet for about ten seconds, while still paused. Long enough that it does not fight the operator while they are actually turning the knob, short enough that they are never left looking at a screen that has forgotten what state it is in.

## Summary

Two fixes (1 and 2), two behaviour changes (3 and 4), plus the `wordReaderAtEof` reset from my previous comment. Everything else in the PR I am happy with — the scoping to `changeSpeed()`/`changeVolume()` rather than the games' value-cores, the self-limiting repaint, and your call to fix `printOnScroll()` at the call site instead of in the library.

I have both devices flashed with your branch, so send the changes over and I will re-test on both.

ckonecny:
Pushed 0f44741, covering all four points:

**1. `getCustomChars()` and the File Player position.** Now saves `fileWordPointer` on entry and restores it on the way out, alongside the `wordReaderAtEof` reset from before. Built and verified this compiles/links cleanly, but I haven't re-tested it on-device this round (ran short on time) - it's the same shape as the already-tested `wordReaderAtEof` fix, but flagging that distinction since you're about to re-test on real hardware.

**2. "No custom set" leftover.** Now clears the line before drawing the message too, not just after - the leftover was visible *during* the message when the previous value (e.g. "LICW Carousel") was longer than "No custom set" itself.

**3. Fallback message.** "No custom set" is now followed by "Fallback: " + whatever `Koch Sequence`'s mapping says for the reverted value (currently "M32", but reads correctly if that target ever changes), each shown for 900ms instead of the previous 700ms total.

**4. Pause prompt reappearing.** Added a `showPausePrompt()` helper that (re-)arms a `pausePromptDeadline` each time it draws "Continue with paddle"; the reappear check lives in the `morseGenerator`/`echoTrainer` loop() cases (`else if (millis() >= pausePromptDeadline) showPausePrompt();`), and the same three sites that already call `MorseOutput::resetTOT()` on activity push the deadline back out again while still paused, so it doesn't fight the operator mid-adjustment.

**Tested on the M32 Pocket:** #2 and #3 together (no leftover character, readable fallback message), and #4 timed at exactly ten seconds with correct re-arming after a further speed change. #1 is build-verified only, as noted above.

PR #214, opened 2026-09-01
Config Tool: Snapshot Export/Import per slot

ckonecny:
## Summary
- Adds **Export** and **Import** buttons to each snapshot slot in the Config Tool's Snapshots tab (Export on stored slots, Import on every slot).
- Export downloads the slot's contents as a JSON file, tagged with the exporting device's hardware/firmware/protocol version.
- Import lets you pick a target slot, warns if the file's firmware/hardware differs from the connected device, and shows a per-parameter File-vs-Current comparison table (same layout as the existing Inspect view) with checkboxes so you can apply everything or just specific values. Confirming applies the checked values to the device and stores the result into the target slot — the same mechanism `Store` already uses (current live settings become the new snapshot content).
- EN + DE user manuals updated to mention export/import alongside the existing set/delete/view/compare wording.

## Possible follow-up (not included here)
The existing Inspect view has never shown a Snapshot-vs-Current comparison for the "Stored Menu" and "Custom Chars" rows — only the snapshot's own value. Import inherits the same gap, since it reuses that same meta display. Adding a "Current" column there is possible (`GET menu` / `GET customchars` both return live state) but touches the pre-existing Inspect feature too, so I left it out of this PR. Happy to add it if you'd like the same parity there.

## Review fixes (commit 0f0133b)
Addressed all six findings from your review:
- **Blocker 1** (silent store failure): the final `put snapshot/store/<n>` now goes through `sendAndParse` and reports failure loudly (log line + `alert()`) instead of discarding the reply.
- **Blocker 2** (no confirmation on an empty slot): import now always confirms, worded to say the live device configuration is being replaced, not just the snapshot.
- **Cross-variant "Stored Menu"**: resolved by matching the file's menu name against this device's own `GET menus` reply; the checkbox is disabled with an explanation when no match is found, instead of applying the file's raw (build-dependent) menu index.
- **Command injection**: parameter name/value and the custom-chars string are validated before being placed in a serial command (name: no `/` or control chars; value: `^\d{1,3}$` and ≤255, matching the uint8_t domain; custom chars: no control chars, `/` still allowed since it's a real Morse character).
- **Pre-checked absent parameters**: rows for parameters not on the connected firmware are now unchecked by default.
- **iOS export no-op**: factored a shared `saveTextFile()` helper (also used by the existing File Builder download) instead of `snapExport()`'s own inline Blob + `<a download>`, and updated `bridge.js` to override that one function instead of just `fbDownload()`.
- Nits: the import confirm button and panel are now included in the disable/hide-on-disconnect handling.

## Test plan
- [x] Christian re-verified live against his Morserino-32 Pocket: the reworded confirm dialog appears on both an empty and an occupied slot (previously the empty-slot case had none), and the disconnect-hide/disable behavior for the import panel.
- [x] Verified live earlier: exported a snapshot with "Font Size" set to Small, switched the device to Normal, imported just that one parameter into an empty slot, and the device switched back to Small instantly.
- [x] Exercised the button layout, comparison table, firmware-mismatch banner, file-rejection path, input-validation (including the exact `\n`-injection payload from the review), cross-variant menu-name resolution/fallback, and the NVS-full store-failure alert — all in a browser session with mocked device replies, not on hardware.
- [ ] Not tested on a Classic M32 — Christian only has a Pocket, so the cross-variant "Stored Menu" fix and the "parameter absent on this firmware" fallback are unverified against a real second variant.
- [ ] Not tested against a real NVS-full condition — the store-failure alert path is simulated only.


oe1wkl (commented):
Thanks Christian — this is a good addition, and a well-judged one. Export/import of snapshots is the obvious missing half of the Snapshots tab, the UI is fitted properly to the tool's existing idiom (the checkbox table reusing the Inspect layout is exactly the right instinct), and the code reads like the rest of the file. Your point about the "Stored Menu"/"Custom Chars" rows lacking a Current column is well taken; agreed it belongs in a separate PR.

I went through it against the firmware and drove the tool in a browser with mocked device replies. Six findings, two of which can lose the user's settings. Everything below is reproduced or read out of the firmware, not inferred.

---

## Blockers

### 1. The final store can fail, and the tool reports success — after the live settings are already gone

`snapImportConfirm()` finishes with

```js
await sendLine('put snapshot/store/'+n);
try{await waitForResponse(2000)}catch(e){}
```

The reply is discarded, so a device *error* is indistinguishable from an OK. And this store really can fail: `doWriteSnapshot()` returns false when NVS is full and the firmware answers

```
{"error":{"name":"error","content":"SNAPSHOT STORE FAILED - NVS FULL?"}}
```

(`m32_v6.ino`, the `type == "snapshot"` / `token == "store"` branch). That is not a theoretical path — snapshots silently failing to store on a full NVS is the bug that produced the whole NVS budget section in `CLAUDE.md` §4, and the reason the firmware now puts a low-space warning on the display at boot.

Reproduced with a stub that answers OK to every `put config` and the error above to the store:

```
device: {"error":{"content":"SNAPSHOT STORE FAILED - NVS FULL?"}}
tool:   [21:14:15] Imported into snapshot 5: 2 applied
```

The user's working configuration has been replaced, no snapshot was written, and nothing tells them. Please use `sendAndParse('put snapshot/store/'+n, ...)` here and report the failure loudly — it is the one step in this flow whose failure the user cannot see for themselves.

(`snapStore()` has the same shape, which is where it was copied from. There it is nearly harmless, because nothing was destroyed first. Here it isn't.)

### 2. Import rewrites the live configuration — with no confirmation at all for an empty slot

The confirm is gated on the slot being occupied:

```js
if(snapExistingList.indexOf(n)!==-1){
  if(!confirm('Import into Snapshot '+n+'? This will overwrite the existing snapshot '+n+'.')) return;
}
```

Verified by instrumenting `sendLine`: importing into an **empty** slot shows no dialog whatsoever and goes straight to `put config/...`, permanently replacing whatever the operator had set up. And when the dialog does appear, it mentions only the snapshot — not the live settings, which are the thing actually being overwritten.

Staging through the live configuration is unavoidable (there is no protocol command that writes snapshot contents directly, and `put snapshot/store` by definition captures the running state), so this is not a design objection. But it has to be stated: confirm on **every** import, and word it for what it does, e.g. *"Import into Snapshot N? This replaces your current settings with the values from the file, and then stores them into snapshot N."*

---

## Should fix before merge

### 3. Cross-variant import silently selects a different mode

`put menu/set/<n>` takes a raw index into `menuText[]`, and that array is conditionally compiled:

- `LORA_DISABLED` (the Pocket) removes the `"LoRa Trx"` entry, shifting everything from index 31 onward by one
- `CONFIG_CW_GAME` adds six entries, `CONFIG_QSO_BOT` more

So index 32 is **WiFi Trx** on a classic M32 and **iCW/Ext Trx** on a Pocket. The import preview shows the file's `menuName` (correct, it is a string), while the import applies the file's `lastExecuted` *number*. The number is in range, `put menu/set` returns OK, and the device lands on a mode the user never saw in the preview. The "Stored Menu" checkbox is checked by default, so this is the default behaviour for exactly the cross-device case the mismatch banner exists to warn about — and the banner only talks about parameters.

Best fix: `GET menus` returns `{content: <path>, "menu number": <n>}` for every entry, so the file's `menuName` can be looked up against the *connected* device and the correct local number used. Cheaper fix: when `device.hardware` differs, uncheck the Stored Menu box by default and say why.

### 4. The imported file can inject arbitrary protocol commands

`c.name`, `c.value` and `snap.customChars.characters` go into the command string unvalidated, and `sendLine(t)` writes `t + '\n'`. A single JSON string value therefore becomes two commands on the wire. Reproduced with `{"name":"word length","value":"5\nput device/reset/defaults"}`:

```
put config/word length/5
put device/reset/defaults
```

The second one factory-resets and reboots the device. It needs a hostile (or merely corrupt) file to trigger, so it is not urgent — but the fix is two regexes: reject a `name` containing `/` or any control character, require `value` to match `^\d{1,3}$` (every config value is a `uint8_t`), and apply the same to the custom-chars string.

### 5. Export is a silent no-op in the iOS app

`Software/iOS/M32Config` loads this very file, unmodified, in a WKWebView. `Resources/Web/bridge.js` already documents this exact failure and overrides `fbDownload()` because of it:

> "Save to Computer" builds a Blob and clicks an `<a download>`. WKWebView ignores that outright: no file, no error, no prompt — and the tool's own log line still says it saved, which is what made this look like a mystery rather than a failure.

`snapExport()` reintroduces the pattern verbatim, log line included, so on the iPhone the Export button does nothing while claiming it saved a file. Given that the tool is deliberately not forked for iOS, the cleanest fix is to factor the blob-save into a single helper —

```js
function saveTextFile(name, text, mime){ ... }   // the existing blob + <a download>
```

— point both `fbDownload()` and `snapExport()` at it, and let `bridge.js` override that one function instead of `fbDownload`. That fixes this export and every future one at the same time; it does mean a matching one-line change in `bridge.js` (and a `sync-webtool.sh` run).

### 6. Parameters this firmware does not have are pre-checked

Rows whose Current column shows `—` are parameters absent from the connected firmware; `put config/<name>/…` will fail on every one of them. Leaving them ticked by default manufactures the failure list the banner then apologises for. Uncheck (or disable) them and note why.

---

## Nits

- `snapImportConfirmBtn` is `class="btn btn-accent"`, not `.btn-sm`, so `enableUI()`'s selector never disables it — it stays live after a disconnect. The import box is not hidden on disconnect either.
- Unchecked rows do not stay "as in the file": since the snapshot is captured from the live device, they are stored with the device's **current** value, so the resulting snapshot is a blend of file and device. Your PR description says so; the UI does not. One line above the table would cover it.
- `.snap-card.empty{opacity:.5}` means the new Import button on an empty slot renders at half opacity and reads as disabled. Store already looks like that, so it is at least consistent — just flagging it.

---

## Everything else checks out

- EN + DE manual wording both updated, and both read correctly.
- No `Software/README.md` change-log entry is owed — config-tool changes have never taken one, so the `sync_whatsnew.py` gate is not in play.
- The card layout survives six buttons: `flex-wrap` gives three rows of two and it still looks tidy.
- Escaping is right — every interpolation of file-supplied text goes through `escHtml()`.
- Hiding the other panel in both directions (Inspect ⇄ Import) was a nice catch.

The eight built manual HTML/PDFs will need regenerating at merge, as usual — that is mine, not yours.

Happy to take another look once the two blockers are addressed.

ckonecny:
Thanks for the thorough review — pushed 0f0133b addressing all six findings (details in the updated PR description). Quick summary:

- Both blockers fixed: the final snapshot store now surfaces failure loudly (log + alert) instead of discarding the reply, and import always confirms now — including on an empty slot — with wording that says the live configuration is being replaced.
- Stored Menu is now resolved by matching the file's menu name against this device's own `GET menus`, and disabled with an explanation when it can't be resolved, instead of trusting the file's raw (build-dependent) index.
- Name/value/custom-chars are validated before hitting the wire — reproduced your `\n`-injection example and confirmed it's rejected.
- Absent-on-this-firmware parameters are unchecked by default now.
- Export and File Builder's download now share one `saveTextFile()` helper, and `bridge.js` overrides that instead of just `fbDownload()`, so Export works correctly in the iOS app too.
- Disconnect now hides/disables the import panel and its confirm button.

Test coverage honestly: Christian re-verified the confirm-dialog and disconnect fixes live on his Pocket. The cross-variant Stored Menu resolution, the absent-parameter fallback, the input validation, and the NVS-full store failure are all exercised only in a browser session with mocked device replies — he only has a Pocket, so the cross-variant and NVS-full paths couldn't be checked against real hardware. Flagging that explicitly rather than claiming more than was actually tested.

oe1wkl (commented):
Went through `0f0133b` and `c228592a`. All six findings are genuinely addressed, and two of the fixes are better than what I asked for. Two smaller issues came in with the fixes, plus two nits — none of them blockers in the way the first two were, so this is your call whether to fold them in here or land them as a follow-up.

Thank you also for the honest test-coverage note. That is exactly the right way to hand over a change, and it told me where to spend my own verification time.

---

## The six, verified

| Finding | Status |
|---|---|
| Store failure swallowed | **Fixed** — `sendAndParse`, plus an `alert()` that says settings changed while the snapshot did not |
| No confirmation on an empty slot | **Fixed** — always confirms, and the wording names the live settings as the thing being replaced |
| Cross-variant menu index | **Fixed properly** — see below |
| Command injection | **Fixed** — I re-ran my `\n` case against your build; rejected |
| Absent-on-this-firmware params pre-checked | Fixed, but see A |
| iOS export no-op | Seam is right, but see B |
| Disconnect handling | **Fixed** — `enableUI` selector, panel hidden, state cleared |

Two are worth calling out properly.

**The menu resolution is exactly right.** Matching `snapshot.menuName` against `GET menus`' `content` works because both sides come out of the same `MorseMenu::getMenuPath()` — so the match is exact by construction, not by coincidence, and it keeps working when the paths are reworded in a later firmware (an old file simply stops matching, and you disable the box with a reason instead of applying a wrong index). Keys check out against `jsonMenuList()`.

**You were right to keep `/` legal in Custom Chars, and I was wrong to lump it in.** `serialDecode` puts everything after the *second* slash into `thirdArg`, so slashes inside the character string arrive intact — and `/` is a real Morse character. Good call, and the comment explaining it will save the next person from "fixing" it.

---

## A. When `GET configs` fails, every parameter is labelled "not on this device"

The unchecked-by-default logic keys off `currentConfigs[c.name]` being absent — but that map is also empty when the *fetch itself* failed, not only when the parameter is genuinely missing from this firmware. `onSnapshotFileSelected` catches that failure, logs one line, and carries on into the preview with `currentConfigs = {}`.

Reproduced against a device answering an error to `get configs`:

```
word length     | not on this device | unchecked
font size       | not on this device | unchecked
colour scheme   | not on this device | unchecked
```

All three exist on the device. The old `—` was at least honest about not knowing; the new label asserts something false, and it does so for the whole table, which reads as "this file is entirely incompatible".

Two consequences worth handling together:

1. The label should distinguish "we could not read the device" from "this firmware does not have it".
2. Import stays clickable in that state. With nothing checked it applies nothing and still runs `put snapshot/store/<n>` — quietly overwriting the target slot with the settings already on the device. Worth blocking the confirm button (or refusing to open the preview at all) when the comparison could not be loaded, since every decision on that screen depends on it.

A single flag set in the `catch` would cover both.

## B. The caller's optimistic success line survives under the new `saveTextFile` seam

The override itself works — I confirmed it reaches calls made from *inside* `snapExport()`, not just inline `onclick` handlers, so the seam does what you intended and future callers get it for free.

But `snapExport()` still logs unconditionally on the line after the call, and so does `fbDownload()`. With the iOS override installed and the save failing, the log reads:

```
Exported snapshot 3 to "morserino_snapshot_3_9.0.0_20260902.json"
Could not save "morserino_snapshot_3_9.0.0_20260902.json": user cancelled
```

Success claimed, then contradicted one line later. That is the failure mode `bridge.js`'s own comment describes — *"the tool's own log line still says it saved, which is what made this look like a mystery rather than a failure"*. The previous override sidestepped it by replacing `fbDownload` wholesale, so the misleading line never ran; the new seam sits *underneath* the caller, which puts it back, now in two places.

Cleanest fix: move the log line into `saveTextFile()` itself — it already has the name and the byte count — and drop it from both callers. The browser implementation logs synchronously, the iOS override logs on resolution, and neither caller can claim success on its own again. Failing that, return the promise and have callers log in `.then()`.

---

## Nits

- `snapImportSetAll()` now skips the Stored Menu checkbox entirely, so **All** will not tick it even when it *was* resolvable. The guard you want is the disabled state, not omission: `if (menuChk && !menuChk.disabled) menuChk.checked = state;`
- The summary counts attempts as successes: `toApply.length + ' applied'` includes the entries that landed in `failed`, so a run reads "42 applied, 2 failed" out of 42 attempted.

---

## The manual rebuild

`c228592a` is correct — exactly the four tracked files, both languages carrying the new wording, and none of the gitignored variant outputs or EPUBs committed by accident. That was on my list to do at merge, so thank you for saving me the step.

Nothing here needs hardware you do not have. A and B both reproduce in a browser session, the way you tested the rest.

ckonecny:
Pushed 63594db addressing A, B, and both nits.

- **A**: added a `snapImportConfigsLoadFailed` flag set when `GET configs` fails. The preview now labels rows "could not verify" (grey) instead of "not on this device" (red) in that case, shows a warning banner explaining why, and disables the Import button — plus a defense-in-depth check at the top of `snapImportConfirm()` itself in case it's ever invoked another way.
- **B**: moved the "Saved ..." log line into `saveTextFile()` itself; `snapExport()` and `fbDownload()` no longer log on their own. Confirmed only one line appears now, win or lose.
- Nit: `snapImportSetAll()` now checks `menuChk.disabled` specifically rather than skipping the checkbox outright, so a resolvable Stored Menu follows All/None again.
- Nit: the import summary now reports `appliedCount` (actual successes) instead of `toApply.length` (attempts), so a partially-failed run no longer claims more successes than it had.

Verified all four in a browser session with mocked device replies (including a `get configs` failure reproducing your exact three-parameter case, and a mixed success/failure batch to check the new count). CI is green on 63594db.

oe1wkl (commented):
`63594db` closes all four. I re-verified each one against your build in a browser session rather than reading the diff, and they behave as described:

- **A** — with `GET configs` failing, the rows now read "could not verify" in grey, the second banner explains why, the Import button is disabled, and calling `snapImportConfirm()` directly still refuses. The distinction between "this firmware does not have it" and "I could not check" is now made in both the label and the colour, which is more than I asked for.
- **B** — exactly one log line in all three paths: browser success, iOS success, iOS cancelled. Moving the message into `saveTextFile()` was the right call over returning a promise; the comment you left there explains *why* the log lives in the helper, which is what stops someone adding a caller-side "saved" line again later.
- **Nit 1** — All/None now drives a resolvable Stored Menu. That test also re-confirmed the cross-variant fix end to end: the file's index 12 was discarded and this device's own number 7 applied.
- **Nit 2** — `2 of 3 parameters applied, 1 failed (colour scheme)`.

The change is good. Two logistics points, neither of them your code:

**There is no CI on this PR.** `gh pr checks 214` reports none. The PR is cross-repo, and `pio-ci.yml` runs `on: [push]`, so it ran in your fork and nothing reports back to this repository. Whatever you saw green is real but invisible here — and more to the point, the two gates that matter for this change (`new-major-version.sh --check` and `sync_whatsnew.py --check`) have not run against the *merge result*, only against your branch. Not something you can fix from your side; noting it so the claim in the PR thread is not taken at face value later.

**The branch now conflicts with master.** Master has moved eleven commits since the branch point (`68d2703`), and both sides touched the same six manual files — the two `.md`, the two generated `.html`, and the two `.pdf`. The generated ones cannot be merged: a text-merge of pandoc HTML is not trustworthy and the PDFs are binary. The recipe is to merge only the Markdown and then rebuild the outputs from it.

So I will take this locally, merge it, rebuild the manuals from the merged sources and push the result, rather than send you into a binary conflict. Your `c228592a` was the right thing to do at the time; it just has to be redone on the other side of the merge. GitHub will still mark this PR merged and credit your commits once they are reachable from master.

Nothing further needed from you. Thank you for the three quick, accurate rounds — and for flagging your test coverage honestly each time; it told me exactly where to spend my own verification.

PR #218, opened 2026-09-06
Sound packs: add "soft & simple" pack

ckonecny:
## Summary

Adds a new sound pack called **soft & simple**: two short, gentle tones — rising for success, falling for error — designed to sit quietly in the background during long sessions without drawing attention.

- `soft & simple/success.mp3` — short, gentle ascending tone
- `soft & simple/error.mp3` — short, gentle descending tone

The README table has been updated to include the new pack with a brief description.

## Files changed

- `Documentation/Sound Packs/soft & simple/success.mp3` _(new)_
- `Documentation/Sound Packs/soft & simple/error.mp3` _(new)_
- `Documentation/Sound Packs/README.md` _(updated table)_


oe1wkl (approved):
Thanks!

oe1wkl:
Christian — the pack is now offered directly in the Configuration Tool (Files → Sound Packs, one click to install), so it is in front of users. Testing it on a Pocket turned up a problem I would like you to have a look at, because the fix is yours to make rather than mine.

**On the built-in speaker the error sound is barely audible.** Not a level problem — a pitch one.

Its dominant frequency is **260 Hz**. The Pocket's loudspeaker radiates almost nothing below about 300 Hz, so most of that sound never leaves the device. High-passing each file at 300 Hz, which is roughly what the speaker actually reproduces, and measuring what is left:

| pack | error, effective level |
|---|---|
| fanfare | −15.5 dB |
| bells | −15.5 dB |
| marimba | −15.7 dB |
| chime | −14.0 dB |
| arcade | −11.5 dB |
| **soft & simple** | **−22.9 dB** |

So it is about 7 dB below every other pack's error — and 7 dB below its own success sound (−16.2 dB), which is why the pair feels lopsided. It loses 6.8 dB through that filter; no other file loses more than 1.6 dB.

Your raw levels are fine, and that is the trap: by peak and RMS the pack is in good shape — its success has the *highest* RMS of any success file in the collection. Turning it up will not help, because the speaker cannot radiate 260 Hz at any level. On headphones you would never hear the problem, which I suspect is how it got through.

**The constraint**, from `devdocs/signal-tones/make_jingles.py`:

> Every error jingle lives at or above ~290 Hz. Below that this speaker radiates almost nothing (see `speaker_mag` in `tone_sim`), so a low, "serious"-sounding error either disappears or has to be scaled past full scale to be heard — which is how the built-in ERR ended up as a clipped square in the first place. Falling gestures in the 300–600 Hz region carry perfectly well and still read as negative.

It is also in `Documentation/Sound Packs/README.md` under "Making your own", third point. Neither is anywhere you would necessarily have looked, and your pack was made outside `make_jingles.py`, so nothing measured it. That is on the documentation as much as on you.

**What would fix it:** bring the error up into roughly 300–600 Hz — an octave up puts it at 520 Hz, comfortably in range, and a falling gesture there still reads as "wrong". Worth looking at the success sound too: at 555 Hz it is the lowest success in the collection (the others run 1045–1570 Hz) and loses 1.1 dB to the same filter, which is probably why Willi described both as on the quiet side. Both files are also the shortest at 0.44 s, and a short sound reads quieter than a sustained one at equal level.

To check your own renders before submitting:

```
ffmpeg -v quiet -i error.mp3 -af "highpass=f=300:poles=2,highpass=f=300:poles=2,astats=measure_overall=RMS_level:measure_perchannel=none" -f null - 2>&1 | grep "RMS level"
```

Compare that against the same measurement without the two `highpass` stages. A difference of more than a decibel or so means the speaker is throwing away sound you meant the user to hear. Aiming the result at roughly −15 dB would put it alongside the rest of the collection.

If you would rather build it inside `make_jingles.py`, the other five packs are defined there as note sequences over partial recipes, and the script measures what it made — that is the path of least surprise, and it means the next person changing the pack gets the same checks.

The pack stays as it is in the meantime: it is perfectly good on headphones, and it is clearly one option among six. No rush, and thank you — the character of it is nice, which is exactly why it is worth getting onto the speaker properly.

